PR #13, opened 2017-03-08
Explain relationship with existing QPP API

marimiyachi:
https://jira.cms.gov/browse/QPPA-365

Very succinctly explained, but can be more verbose if you want @lucasmbrown-usds. My plan next is to link from the developer docs just explaining what qpp-measures-data is, but not mention the relationship with the existing QPP API in that text. LMK what you think.

Reviewer: @kencheeto (I spellchecked myself this time :P)

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,14 +1,18 @@
 # qpp-measures-data
-Quality Payments Program Measures Data Repository
+Quality Payment Program Measures Data Repository
 
 This repository hosts measures data for QPP and supports functionality to import
 measures data as an NPM module. It currently hosts all Improvement Activity and
 Advancing Care Information measures, as well as a fraction of Quality measures.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,10 +1,19 @@
 # qpp-measures-data
-Quality Payments Program Measures Data Repository
+Quality Payment Program Measures Data Repository
```
kencheeto: maybe we can use this line as the header?

marimiyachi: sgtm!

lucasmbrown-usds: also @marimiyachi @kencheeto i don't think we ever implemented this comment?

marimiyachi: I didn't push the commit but if you look at the README the change was made :)

PR #14, opened 2017-03-15
QPPA-320/stub benchmarks

sungchoi:
# QPPA-350/stub benchmarks
### Notes:

JIRA [#QPPA-320](https://jira.cms.gov/browse/QPPA-320)
[Confluence Tech Specs Doc](https://docs.google.com/document/d/1M9GSOviKleVAECwOP8l1H-vpkpGdXiTCe8vwFhGbwkQ)
[Deciles Example Scenarios](https://docs.google.com/spreadsheets/d/1quaCAnQLfJ-Z58AKOFsLxW18kc0IlRNFvTj2FR6kqrA)

### Usage:

- Run `node scripts/parse-2016-benchmarks-data.js` in order to regenerate `benchmarks/2016.json`
- `getBenchmarksData` and `getBenchmarksSchema` functions are available via the module (index.js)

### Work remaining:

- [x] Squash commits.
- [x] Rerun the `convert-qpp-to-measures` script again in order to add actual `isInverse` values.
- [x] Add entry point functions `getBenchmarksData` and `getBenchmarksSchema` into index.js
- [x] Reasonable test coverage.
- [x] Update measure objects' `isInverse` property with actual data from benchmarks
    - We can either
        - add an `isinverse` property to the benchmarks in its json file and use that as the source of info.
        - OR we can update the `measures-data.json` and `xml` inside of the benchmark parsing script.
        - -> OR we can update the `convert-qpp-to-measures` script to add `isInverse` value during its execution.
    - There is a `isInverseBenchmark` which expects the csv record object as its argument, but we could make one that expected the parsed and formatted benchmark object.
- [x] Validate updated measures-schema.yaml
    - [x] I think it would be nice to split out the benchmark schema and then imported into the measures-schema.yaml. Is that possible?
- Update the submissions-api repo to use the benchmark data.
- [x] QPPA-350: verify that we're good to go with this data structure and that the benchmark-data.xls we are using is correct. If so, we should be able to use this parsed data as the 2016 historical benchmark data.

### Automated Test Plan

- [ ] `npm test` should be green
- [ ] `cat benchmarks/2016.json | node util/validate-data.js benchmarks` should come back `Valid!`

### Manual Test Plan
 
- [ ] Code review


util/constants/quality-id-to-measure-map.js
```diff
@@ -0,0 +1,22 @@
...
+var keyBy = require('lodash/keyBy');
+// Data
+var measures = require('../../measures/measures-data.json');
+
+/**
+ * @type {{}} - mapping of integer qualityIds to corresponding measure
+ */
+var QUALITY_ID_TO_MEASURE_MAP = keyBy(measures, function(measure) {
+  /**
+   * NOTE(sung): The qualityId is usually a string integer.
```
sungchoi: I can probably fix this by just removing leading zeros instead of using `parseInt`.

data/historical-benchmarks/demo.csv
```diff
@@ -0,0 +1,689 @@
+Table 2: MIPS Benchmark Results ,,,,,,,,,,,,,
```
abarciauskas-bgse: Wondering if this is just for demo purposes it makes sense to store this whole file. I think it's good to have an example file for testing purposes but maybe we want to just have a few of each scenario we want to cover in testing. Of course we'll generate the benchmarks using all the data but don't think all this original data needs to be stored here.

measures/measures-schema.yaml
```diff
@@ -28,6 +28,8 @@ definitions:
         description: ACI measures can belong to the transition measure set. Quality measures can belong to multiple measure sets that represent different specialties.
         type: array
         items: { $ref: #/definitions/measureSets }
+      benchmark:
```
abarciauskas-bgse: I may have missed a follow up conversation with Mari but I'm not sure we want to store benchmarks nested in measures since measures can be multi year but benchmarks can only represent a single year.

test/index.js
```diff
@@ -1 +1,5 @@
 require('./measures/measures-data-spec');
+require('./util/is-inverse-benchmark-spec');
+require('./util/format-decile-generator-spec');
+require('./util/format-csv-record-spec');
+require('./benchmarks/benchmarks-data-spec');
```
abarciauskas-bgse: Hmm should `test/benchmarks/benchmarks-data-spec.js` be removed? Doesn't seem to be anything there. Or if there is an intention to add tests maybe we want to add a file with a comment to that effect.

util/format-decile-generator.js
```diff
@@ -0,0 +1,77 @@
...
+  return function(decileString, index, array) {
+    var range           =  decileString ? decileString.match(/([0-9]*[.]?[0-9]+)/g) : null;
+    var nextIndex       = index + 1;
+    var nextDecile      = array[index + 1];
+    var nextDecileRange = nextDecile ? nextDecile.match(/([0-9]*[.]?[0-9]+)/g) : null;
+    // if self is undefined and a previous decile is defined, return null.
+    // if self is undefined and no previous decile is defined, but nextDecile is defined, return nextDecile's lowerbound.
+    // if self is undefined, and the nextDecile is undefined, return null.
+    // if self is defined, but nextDecile is not, find the next defined decile above.
+    // if self is defined and self is Decile 10 return the explicit upper bound or top.
```
marimiyachi: I know this is rehashing previous conversations....but why not use the explicit lower bound as well?

util/format-csv-record.js
```diff
@@ -0,0 +1,68 @@
...
+ *  performanceYear: string,
+ *  submissionMethod: string,
+ *  deciles: Array} | undefined} - benchmark object
+ */
+var formatCsvRecord = function formatCsvRecord(record, options) {
+  /**
+   * NOTE: Some of the benchmarks don't correspond to
+   * any of the measures currently in our json.
+   */
+  var measure = QUALITY_ID_TO_MEASURE_MAP[record.qualityId] || {};
```
marimiyachi: So if a measure doesn't exist for the benchmark it's still being added will a null measureId? Don't think that's the behavior we want. I'd say don't worry that we have benchmarks that don't correspond to measures - the measure could be deprecated or it may be a non-QPP measure (which hasn't been finalized yet).

scripts/parse-demo-benchmarks-data.js
```diff
@@ -0,0 +1,31 @@
...
+parse(benchmarksData, {columns: DEMO_CSV_COLUMNS, from: 4}, function(err, records) {
+  if (err) {
+    console.log(err);
+  } else {
+    records.forEach(function(record) {
+      var qualityId = record.qualityId;
+
+      // NOTE: qualityId is being used here since not all benchmarks currently have corresponding measures.
+      // This will prevent the loss of benchmarks during parsing.
+      if (!benchmarks[qualityId]) benchmarks[record.qualityId] = {};
```
marimiyachi: I'm confused why you're renaming the measure ID field to be quality ID?

benchmarks/demo.json
```diff
@@ -0,0 +1 @@
...00]}},"GIQIC 2":{"qcdrOrQualifiedRegistry":{"benchmarkYear":"2016","performanceYear":"2018","submissionMethod":"qcdrOrQualifiedRegistry","deciles":[null,91.43,93.73,95.57,96.65,97.47,98.12,98.68,99.28,100]}},"GIQIC 3":{"qcdrOrQualifiedRegistry":{"benchmarkYear":"2016","performanceYear":"2018","submissionMethod":"qcdrOrQualifiedRegistry","deciles":[null,95.36,96.83,97.5,98.03,98.51,98.86,99.23,99.58,100]}},"GIQIC 4":{"qcdrOrQualifiedRegistry":{"benchmarkYear":"2016","performanceYear":"2018","submissionMethod":"qcdrOrQualifiedRegistry","deciles":[null,96.46,97.62,98.31,98.77,99.11,99.44,99.69,100,100]}},"GIQIC 5":{"qcdrOrQualifiedRegistry":{"benchmarkYear":"2016","performanceYear":"2018","submissionMethod":"qcdrOrQualifiedRegistry","deciles":[null,null,null,null,null,null,null,null,0,0]}},"GIQIC 6":{"qcdrOrQualifiedRegistry":{"benchmarkYear":"2016","performanceYear":"2018","submissionMethod":"qcdrOrQualifiedRegistry","deciles":[null,82.14,87.01,89.66,91.84,93.75,95.1,96.6,98.21,100]}},"GIQIC 8":{"qcdrOrQualifiedRegistry":{"benchmarkYear":"2016","performanceYear":"2018","submissionMethod":"qcdrOrQualifiedRegistry","deciles":[null,null,null,null,null,null,null,null,0,0]}},"GIQIC 9":{"q
```
sungchoi: @marimiyachi By flat do you mean just an array of benchmark objects? Not a problem. I just figured this format would make it easier when we have to match the measures with the benchmarks.
```
{
  [qualityId]: {
    [submissionType]: Benchmark
}
```

test/util/format-csv-record-spec.js
```diff
@@ -0,0 +1,237 @@
...
+var expect  = chai.expect;
+// Utils
+var formatCsvRecord = require('./../../util/format-csv-record');
+
+var options = {
+  benchmarkYear: '2016',
+  performanceYear: '2018'
+};
+
+describe('formatCsvRecord', function() {
```
sungchoi: Could still use a little work, but Thank You!! ^.^

util/format-decile-generator.js
```diff
@@ -0,0 +1,77 @@
...
+
+  /**
+   * Params correspond to the Array.map signature
+   * @param {string?} decileString - from csv
+   * @param {number} index
+   * @param {Array} array
+   * @return {number | null}
+   */
+  return function(decileString, index, array) {
+    var range           =  decileString ? decileString.match(/([0-9]*[.]?[0-9]+)/g) : null;
```
sungchoi: Will do!

util/format-decile-generator.js
sungchoi: @marimiyachi Yea, I had similar thoughts as I was writing this. I think we might have over optimized. Might be better to just copy over what was given in the xls, but we still have a few issues/decisions:
1) If Decile 1 is "0-1.02" and Decile 2 is "1.03-2.00", what do we do for values that are less than 1.03 but greater than 1.02 (1.02 < x < 1.03 )? 
2) Still need to parse the Deciles that aren't in "n-m" format, e.g. "<= x", "100", and "0". 

Also we should ask, what will be easiest for when we are the ones maintaining the data and spitting out benchmarks? 

Maybe this calls for a 10 minute chat.

util/format-csv-record.js
sungchoi: Good to know. I will update. I also want to double check to make sure that the measureId we were given in the xls sheet doesn't sometimes correspond to something other than measures' qualityId, which is definitely the case for the a majority of the benchmarks. 

@marimiyachi  **Question:** Will it be possible for the set of Measures that exist to change? If so, will we need/want to regenerate the benchmarks whenever measures are updated?

scripts/parse-demo-benchmarks-data.js
sungchoi: the `measureId` in the xls sheet seems to match the `qualityId` for a measure object, not it's `id`.

test/index.js
sungchoi: Haven't quite decided if there will be tests there, but will clean up and remove if it remains empty.

measures/measures-schema.yaml
sungchoi: 👍

benchmarks/demo.json
marimiyachi: I mean like `[{ "measureId": "1", "submissionMethod": "ehr", ...}, { "measureId": "2", "submissionMethod": "ehr", ...}]`. It's redundant to have the measureId and submissionMethod in the value of the map, but it's nice to make it explicit. Also nit to pretty print this into the output file.

util/format-decile-generator.js
marimiyachi: Yes - let's chat after standup!

util/format-csv-record.js
marimiyachi: Yeah let's definitely double check about quality ID. The set of measures will change from year to year, which is why we embed `performanceYear`.

scripts/parse-demo-benchmarks-data.js
marimiyachi: 😭

marimiyachi: We probably want to get confirmation - maybe Steven Szeliga can help.

util/format-csv-record.js
```diff
@@ -0,0 +1,73 @@
...
+var formatCsvRecord = function formatCsvRecord(record, options) {
+  /**
+   * NOTE: Some of the benchmarks don't correspond to
+   * any of the measures currently in our json.
+   */
+  var measure = QUALITY_ID_TO_MEASURE_MAP[record.qualityId];
+
+  if (!measure) return;
+  // TODO(sung): Double check that we can/should leave these out.
+  if (record.benchmark.trim() === 'N') return;
```
marimiyachi: Yeah - I think you're all good re the comment.

util/constants/quality-id-to-measure-map.js
```diff
@@ -0,0 +1,18 @@
+// Libraries
+var keyBy = require('lodash/keyBy');
+// Data
+var measures = require('../../measures/measures-data.json');
+
+/**
+ * @type {{}} - mapping of integer qualityIds to corresponding measure
+ */
+var QUALITY_ID_TO_MEASURE_MAP = keyBy(measures, function(measure) {
```
marimiyachi: Rather than put this in a separate file can you put it in `format-csv-records`? Since it is only used there.

scripts/parse-2016-benchmarks-data.js
```diff
@@ -0,0 +1,40 @@
+// Libraries
+var fs    = require('fs');
```
marimiyachi: Update name to maybe just `parse-benchmarks-data`?

util/constants/demo-csv-columns.js
```diff
@@ -0,0 +1,18 @@
+var DEMO_CSV_COLUMNS = [
```
marimiyachi: I'd say same is true here - this can live in the parse benchmarks script.

util/constants/submission-method-map.js
```diff
@@ -0,0 +1,16 @@
+/**
+ * Maps normalized (trimmed and squeezed) submission method values
+ * from csv to standard values.
+ * @enum {string}
+ */
+var submissionMethodMap = {
```
marimiyachi: Not used anywhere?

util/format-csv-record.js
```diff
@@ -0,0 +1,73 @@
...
+ * @returns {{
+ *  measureId: string,
+ *  benchmarkYear: string,
+ *  performanceYear: string,
+ *  submissionMethod: string,
+ *  deciles: Array<null|number>
+ *  } | undefined
+ * } - benchmark object
+ */
+var formatCsvRecord = function formatCsvRecord(record, options) {
```
marimiyachi: Maybe make this name more benchmark specific? `FormatBenchmarkRecord`?

util/format-csv-record.js
```diff
@@ -0,0 +1,73 @@
...
+      record.decile3,
+      record.decile4,
+      record.decile5,
+      record.decile6,
+      record.decile7,
+      record.decile8,
+      record.decile9,
+      record.decile10
+    ]
+      .map(formatDecileGenerator(record))
```
marimiyachi: I think a general rule of thumb for determining if a function should be broken out into a util class is to answer if it will be reused. In this case, both `formatSubmissionMethod` and `formatDecileGenerator` are providing very specific functionality to this parent function, but aren't used anywhere else and don't have a clear use case of how/why they would be - so I'd keep them in this file.

scripts/parse-2016-benchmarks-data.js
```diff
@@ -0,0 +1,40 @@
...
+      var isInverse = isInverseBenchmarkRecord(record);
+
+      if (benchmark) {
+        benchmarks.push(benchmark);
+        idToIsInverseMap[benchmark.measureId] = isInverse;
+      }
+    });
+
+    fs.writeFileSync(path.join(__dirname, '../benchmarks/2016.json'), JSON.stringify(benchmarks, null, 2), 'utf8');
+    fs.writeFileSync(path.join(__dirname, '../benchmarks/measure-id-to-is-inverse-map.json'), JSON.stringify(idToIsInverseMap, null, 2), 'utf8');
```
marimiyachi: Hm....having this intermediate file is not ideal since this doesn't have much value on it's own. Maybe instead bring the benchmarks JSON into `convert-qpp-to-measures` and refactor `isInverseBenchmarkRecord` to take the benchmark JSON as an argument instead of the csv row?

util/is-inverse-benchmark.js
```diff
@@ -0,0 +1,18 @@
+/**
+ *
+ * @param {{}} benchmark
+ * @returns {boolean}
+ */
+var isInverseBenchmark = function isInverseBenchmark(benchmark) {
```
marimiyachi: Oh nice you already have this! Yeah - I'd suggest using this function in `convert-qpp-to-measures` instead of persisting the map.

util/convert-qpp-to-measures.js
```diff
@@ -179,15 +184,16 @@ function parseQpp(json) {
       } else if (category === 'quality') {
         // TODO (Mari): These will need to be populated via another mechanism
         // outside of the API (similar to CEHRT eligible IA measures)
-        obj.isInverse = false;
+        // isInverse defaults to false;
```
sungchoi: 👍

util/constants/demo-csv-columns.js
sungchoi: 👍

scripts/parse-2016-benchmarks-data.js
sungchoi: @marimiyachi There are cases in which we won't be able to tell if it's direct or inverse after we've parsed it into the 9 float decile list. Namely [0, 0, 0, 0, 0, 0, 0, 0, 0] and [100, 100, 100, 100, 100, 100, 100, 100, 100]. But, I will try to see if I can clean this up more in another way.

I think maybe the simpliest thing would be to just remove this line, since we've already generated the MAP, assuming that no benchmarks will every flip from direct to inverse or vice versa. This would also assume that the set of benchmarks would never expand and that this file would always include the whole set of measures that need to know whether they are direct or inverse. hmmm.

scripts/parse-2016-benchmarks-data.js
```diff
@@ -0,0 +1,40 @@
...
+/**
+ *
+ * Script to generate benchmark.json file from csv
+ * To run: `node parse-2016-benchmarks-data.js`
+ *
+ * TODO(sung): Generalize this script. Make it take commandline arguments for the year.
+ */
+var benchmarks = [];
+var idToIsInverseMap = {};
+
```
sungchoi: 👍

scripts/parse-2016-benchmarks-data.js
sungchoi: 👍

index.js
```diff
@@ -1,12 +1,27 @@
...
 var YAML = require('yamljs');
 
+/**
+ *
+ * @param {number|string} year - four digit year
+ * @return {Array<Benchmark>} - benchmarks data for given year
+ */
+exports.getBenchmarksData = function(year) {
+  return JSON.parse(
+    fs.readFileSync(path.join(__dirname, 'benchmarks', year + '.json')));
```
sungchoi: 👍

benchmarks/benchmarks-schema.yaml
```diff
@@ -0,0 +1,23 @@
...
+items: { $ref: #/definitions/benchmark }
+
+definitions:
+  benchmark:
+    type: object
+    properties:
+      measureId:
+        description: Corresponds to measure object's measureId.
+        type: string
+      performanceYear: { type: string }
```
sungchoi: 👍

test/util/is-inverse-benchmark-record-spec.js
```diff
@@ -0,0 +1,122 @@
...
+    });
+
+    expect(isInverse2, 'case2').to.be.true;
+  });
+  it('should return false when decile 10 is >= x', function() {
+    var isInverse = isInverseBenchmarkRecord({decile10: '>= 5.00'});
+
+    expect(isInverse).to.be.false;
+
+    var isInverse2 = isInverseBenchmarkRecord({
```
sungchoi: Forgot to finish writing that test. WompWomp

util/format-csv-record.js
sungchoi: 👍

util/format-csv-record.js
sungchoi: 👍

benchmarks/benchmarks-schema.yaml
```diff
@@ -0,0 +1,28 @@
...
+    type: object
+    properties:
+      measureId:
+        description: Corresponds to measure object's measureId.
+        type: string
+      performanceYear:
+        description: The year for which data is being submitted
+        type: string
+      benchmarkYear:
+        description: The year against which the data is be compared.
```
abarciauskas-bgse: Grammar: `is being compared`

benchmarks/benchmarks-schema.yaml
```diff
@@ -0,0 +1,28 @@
...
+        type: string
+      deciles:
+        description: List of 9 numbers representing the inclusive lower bound of Deciles 2 through 10. The top and bottom are implied to be 100 and 0 respectively for direct measures and 0 and 100 for inverse measures. The range of any given decile begins at it's lower bound and continues up to but does not include the subsequent decile's lower bound. If the subsequent decile's lower bound is equal to the current decile's lower bound, then that decile is undefined or, in other words, empty.
```
abarciauskas-bgse: Suggested revised version:
```
The deciles list object is a list of 9 numbers representing deciles of measurement values for a given measure, performance year and submission method. The 9 numbers represent the inclusive lower bounds of deciles 2 through 10. The upper and lower bounds of the measurement value range are implied to be 100 and 0 respectively for direct measures and 0 and 100 respectively for inverse measures. The range of any given decile begins at its lower bound and continues up to but does not include the subsequent decile's lower bound. If the subsequent decile's lower bound is equal to the current decile's lower bound, then that decile is undefined or, in other words, empty.
```

Changes:
* uncapitalized Deciles in the first sentence
* Added summary sentence at the beginning
* Replace top and bottom with upper and lower bounds
* Replace "it's lower bound" with "its lower bound"

benchmarks/measure-id-to-is-inverse-map.json
```diff
@@ -0,0 +1,134 @@
+{
```
abarciauskas-bgse: Nice

benchmarks/benchmarks-schema.yaml
sungchoi: 👍

scripts/parse-benchmarks-data.js
```diff
@@ -0,0 +1,76 @@
...
+      }
+    });
+});
+} else {
+  console.log('Please provide a performance year like so: `node parse-benchmarks-data.js 2018`');
+}
+
+
+
+
```
abarciauskas-bgse: Nit: remove extra lines

util/is-inverse-benchmark-record.js
```diff
@@ -0,0 +1,47 @@
+/**
+ *
+ * @param {{}} record
+ * @return {boolean}
+ */
+var isInverseBenchmarkRecord = function isInverseBenchmarkRecord(record) {
```
abarciauskas-bgse: Would defer to @marimiyachi but the pattern in this repo for function declarations to omit the var declaration, e.g.:

```javascript
function isInverseBenchmarkRecord(record) {
  ...
}
```

measures/measures-data.json
```diff
@@ -3097,7 +3097,7 @@
     "nqfId": null,
     "qualityId": "378",
     "isHighPriority": true,
-    "isInverse": true,
+    "isInverse": false,
```
sungchoi: This should be true :(

scripts/convert-qpp-to-measures.js
```diff
@@ -42,6 +36,30 @@ var CEHRT_ELIGABLE_IA_MEASURE_IDS = [
...
+var benchmarksData   = fs.readFileSync(path.join(__dirname, './../data/historical-benchmarks/2015.csv'), 'utf8');
+var benchmarkRecords = parse(benchmarksData, {columns: BENCHMARK_CSV_COLUMNS, from: 4});
+// Commandline arguments
+var category = process.argv[2] || 'ia';
+// Variables
+var qualityIdToIsInverseMap = {};
+var qpp = '';
+
+benchmarkRecords.forEach(function(record) {
+  // NOTE: There is a special case, e.g. qualityId 378 EHR, which has decile
```
sungchoi: @marimiyachi Figured it out. There was a conflict between EHR and Registry/QCDR versions of 378. As for qualityId 1, it was because '001' does not equal '1'. I had to normalize the id for correct lookup in the map.

measures/measures-data.json
```diff
@@ -1362,7 +1362,7 @@
...
+    "description": "Participating in a Rural Health Clinic (RHC), Indian Health Service (IHS), or Federally Qualified Health Center in ongoing engagement activities that contribute to more formal quality reporting, and that include receiving quality data back for broader quality improvement and benchmarking improvement which will ultimately benefit patients. Participation in Indian Health Service, as an improvement activity, requires MIPS eligible clinicians and groups to deliver care to federally recognized American Indian and Alaska Native populations in the U.S. and in the course of that care implement continuous clinical practice improvement including reporting data on quality of services being provided and receiving feedback to make improvements over time.",
```
sungchoi: @marimiyachi Looks like there actually was a small description and title update from the cms endpoints we curl from.

marimiyachi: 👍

util/is-inverse-benchmark-record.js
marimiyachi: Hm...I see that I did that elsewhere but I'm not sure what the best convention is. I'd say let's at least get rid of the repetition here so `var isInverseBenchmarkRecord = function(record) {...}`

scripts/convert-qpp-to-measures.js
marimiyachi: Oh geez. So quality IDs are not unique? I'd explain this a little more because right now it seems dependent on order - explain that all benchmarks for a particular measure will have the same isInverse flag, some benchmark records will have no data so the flag will be false and in these cases when we come up with a benchmark with the same quality id, we always want to double check.

scripts/parse-benchmarks-data.js
```diff
@@ -0,0 +1,65 @@
...
+ * Script to generate benchmark.json file from csv
+ * To run: `cat [DATA_CSV_FILE] | node scripts/parse-benchmarks-data.js [BENCHMARK_YEAR] [PERFORMANCE_YEAR]`
+ * e.g. `cat data/historical-benchmarks/2015.csv | node scripts/parse-benchmarks-data.js 2015 2017`
+ */
+var benchmarks = [];
+// Commandline Arguments
+var benchmarkYear   = process.argv[2];
+var performanceYear = process.argv[3];
+
+if (performanceYear) {
```
sungchoi: 👍

scripts/convert-qpp-to-measures.js
sungchoi: 👍

README.md
```diff
@@ -7,39 +7,53 @@ Advancing Care Information measures, as well as a fraction of Quality measures.
...
+(benchmark year 2015).
+
 ### Importing the qpp-measures-data module
 Functions take a string version argument and return the appropriate YAML schema or JSON data.
 The module can be used with the following pattern:
 ```javascript
 var qppMeasuresData = require('@cmsgov/qpp-measures-data');
 var measuresData = qppMeasuresData.getMeasuresData();
 var measuresSchema = qppMeasuresData.getMeasuresSchema();
+var benchmarksData = qppMeasuresData.getBenchmarksData();
```
sungchoi: 👍

util/is-inverse-benchmark-record.js
sungchoi: I prefer using the named functions to help with debugging, but I've noticed that a lot of the newer environments seem to name anonymous functions for you based on their variable name or make it seem like they're named while you're debugging. So, it's probably not as compelling of a reason nowadays. 

Also, prefer putting function expressions up top over the declarations that hoist and are usually placed below where they're called, but I'm not overly opinionated on this. When in Rome.

gabesmed: yes -- node is smart enough to use the name of the var in stack traces, so there is no longer any reason to name the function twice. either name the var, or the function

scripts/parse-benchmarks-data.js
```diff
@@ -0,0 +1,65 @@
...
+ * Script to generate benchmark.json file from csv
+ * To run: `cat [DATA_CSV_FILE] | node scripts/parse-benchmarks-data.js [BENCHMARK_YEAR] [PERFORMANCE_YEAR]`
+ * e.g. `cat data/historical-benchmarks/2015.csv | node scripts/parse-benchmarks-data.js 2015 2017`
+ */
+var benchmarks = [];
+// Commandline Arguments
+var benchmarkYear   = process.argv[2];
+var performanceYear = process.argv[3];
+
+if (benchmarkYear && performanceYear) {
```
sungchoi: Updated to check for benchmarkYear

README.md
```diff
@@ -7,39 +7,53 @@ Advancing Care Information measures, as well as a fraction of Quality measures.
...
+(benchmark year 2015).
+
 ### Importing the qpp-measures-data module
 Functions take a string version argument and return the appropriate YAML schema or JSON data.
 The module can be used with the following pattern:
 ```javascript
 var qppMeasuresData = require('@cmsgov/qpp-measures-data');
 var measuresData = qppMeasuresData.getMeasuresData();
 var measuresSchema = qppMeasuresData.getMeasuresSchema();
+var benchmarksData = qppMeasuresData.getBenchmarksData(2017);
```
sungchoi: Added year argument to example

scripts/convert-qpp-to-measures.js
```diff
@@ -42,6 +36,34 @@ var CEHRT_ELIGABLE_IA_MEASURE_IDS = [
...
+var benchmarksData   = fs.readFileSync(path.join(__dirname, './../data/historical-benchmarks/2015.csv'), 'utf8');
+var benchmarkRecords = parse(benchmarksData, {columns: BENCHMARK_CSV_COLUMNS, from: 4});
+// Commandline arguments
+var category = process.argv[2] || 'ia';
+// Variables
+var qualityIdToIsInverseMap = {};
+var qpp = '';
+
+benchmarkRecords.forEach(function(record) {
+  // NOTE: qualityId is not unique per record. (A qualityId and submissionMethod
```
sungchoi: Updated comment.

util/format-benchmark-record.js
```diff
@@ -0,0 +1,189 @@
...
+  return measure.qualityId ? measure.qualityId.replace(/^0*/, '') : undefined;
+});
+// Helper Functions
+/**
+ *
+ * @param {string} submissionMethod - non-normalized version from csv dataset
+ * @returns {string} - normalized version
+ */
+var formatSubmissionMethod = function(submissionMethod) {
+  return SUBMISSION_METHOD_MAP[submissionMethod.replace(/\s/g, '').toLowerCase()];
```
sungchoi: removed function names from function expressions.

sungchoi:
Back to you @marimiyachi

scripts/convert-qpp-to-measures.js
marimiyachi: 👍

scripts/parse-benchmarks-data.js
```diff
@@ -0,0 +1,65 @@
...
+
+          if (benchmark) benchmarks.push(benchmark);
+        });
+
+        fs.writeFileSync(path.join(__dirname, '../benchmarks/' + performanceYear + '.json'), JSON.stringify(benchmarks, null, 2), 'utf8');
+      }
+    });
+});
+} else {
+  console.log('Please provide a performance year like so: `node parse-benchmarks-data.js 2018`');
```
marimiyachi: Update this error message to mention benchmarkYear as well

PR #16, opened 2017-03-24
Updates package.json. Adds csv-parse to dependencies.

sungchoi:
Realized that I didn't check csv-parse package into package.json :(

### DIFF
--- a/index.js
+++ b/index.js
@@ -3,14 +3,25 @@ var fs = require('fs');
 var path = require('path');
 var YAML = require('yamljs');
 
+var yearRegEx = /^[0-9]{4}/;
+var benchmarkJsonFileRegEx = /^[0-9]{4}\.json$/;
+
 /**
  *
- * @param {number|string} performanceYear - four digit year
- * @return {Array<Benchmark>} - benchmarks data for given year
+ * @return {{}} - benchmarks data -
+ * An object keyed by performance year with array values
+ * containing the benchmarks for that performance year
  */
-exports.getBenchmarksData = function(performanceYear) {
-  return JSON.parse(
-    fs.readFileSync(path.join(__dirname, 'benchmarks', performanceYear + '.json')));
+exports.getBenchmarksData = function() {
+  var benchmarksByYear = {};
+
+  fs.readdirSync(path.join(__dirname, 'benchmarks')).forEach(function(file) {
+    if (benchmarkJsonFileRegEx.test(file)) {
+      benchmarksByYear[file.match(yearRegEx)[0]] = require(path.join(__dirname, 'benchmarks', file));
+    }
+  });
+
+  return benchmarksByYear;
 };
 
 /**
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@cmsgov/qpp-measures-data",
-  "version": "0.0.21",
+  "version": "0.0.22",
   "description": "Quality Payment Program Measures Data Repository",
   "repository": {
     "type": "git",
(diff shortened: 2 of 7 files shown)


marimiyachi:
oops! lgtm

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@cmsgov/qpp-measures-data",
-  "version": "0.0.21",
+  "version": "0.0.22",
```
kencheeto: are we publishing to npm?

gabesmed: yes!

kencheeto: ach, I didn't realize we were publishing this privately (since it's open source), explains some issues I had with yarn work!

sungchoi:
@marimiyachi @kencheeto Could you take a look at this again. I added another small commit that makes `getBenchmarksData` return undefined instead of throwing an error when the file (data for a given performanceYear) doesn't exist.

index.js
```diff
@@ -6,11 +6,15 @@ var YAML = require('yamljs');
...
+ * @return {Array<Benchmark>|undefined} - benchmarks data for given year
  */
 exports.getBenchmarksData = function(performanceYear) {
-  return JSON.parse(
-    fs.readFileSync(path.join(__dirname, 'benchmarks', performanceYear + '.json')));
+  try {
+    return JSON.parse(
+      fs.readFileSync(path.join(__dirname, 'benchmarks', performanceYear + '.json')));
+  } catch (error) {
+    return;
```
sungchoi: Not sure if I should check the error code here. Other than the file not existing, I'm not sure what errors exist and if/how they should be handled.

kencheeto: If this is an unrecoverable error, as a library, we should throw a documented error and let the application handle the failure.

Tangentially, is this the first PR that introduces a dependency on `fs`? Do we need to make that clear somewhere? If there's value in a serverless use of this library, maybe we should consider a build step to bake benchmarks into the exported file.

sungchoi: Am I reading this correctly? 
Your first suggestion is to leave the try/catch out like it was before and just document the fact that that may happen and that anyone using this module should handle that on their end?

Your second suggestion is to add a build task that aggregates all of the historical benchmarks into one file (index file) that lives in the same place as this function so that we don't have to use `fs` in the first place.

Is that right? @kencheeto

kencheeto: @sungchoi i just educated myself on the repo and see that we're already loading files from fs. I think to be consistent, we could:

1. catch file-not-found (`catch(error) { error.code === 'ENOENT'`) errors and rethrow with a more descriptive message like 'no benchmark found for year x'
2. rethrow all other errors

re: second suggestion, looks like we've made the decision to load from fs already, so I think that's fine to stick with!

what do you think?

sungchoi:
@marimiyachi @kencheeto Can you take another look at this? I updated the getBenchmarksData method and updated the script and resulting files to have performanceYear and benchmarkYear be integer numbers instead of strings.

index.js
```diff
@@ -3,14 +3,25 @@ var fs = require('fs');
...
+ * containing the benchmarks for that performance year
  */
-exports.getBenchmarksData = function(performanceYear) {
-  return JSON.parse(
-    fs.readFileSync(path.join(__dirname, 'benchmarks', performanceYear + '.json')));
+exports.getBenchmarksData = function() {
+  var benchmarksByYear = {};
+
+  fs.readdirSync(path.join(__dirname, 'benchmarks')).forEach(function(file) {
+    if (jsonRegEx.test(file)) {
```
marimiyachi: Instead of a regex could you parse the file and then do a `keyBy`?

sungchoi: Sorry I don't understand this comment. :(

I'm trying to skip over the yaml file with this if statement, and a keyBy seems overkill since we know which year each collection belongs to without iterating over them.

sungchoi: oh I see. Instead of using the regex to get the year.

sungchoi: still. seems like overkill to iterate through the entire collection.

marimiyachi: you're already iterating over the entire collection though?

PR #19, opened 2017-04-11
Update qcdr to registry for benchmark data

marimiyachi:
https://jira.cms.gov/browse/QPPA-491

Updates `qcdrOrQualifiedRegistry` to `registry` - practically speaking they are the same. 

Testing
`npm run test` passes
`cat benchmarks/2017.json | node scripts/validate-data.js benchmarks` passes

### DIFF
--- a/util/format-benchmark-record.js
+++ b/util/format-benchmark-record.js
@@ -14,7 +14,7 @@ var measures = require('./../measures/measures-data.json');
 var SUBMISSION_METHOD_MAP = {
   'claims': 'claims',
   'registry': 'registry',
-  'registry/qcdr': 'qcdrOrQualifiedRegistry',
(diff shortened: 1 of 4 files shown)


util/format-benchmark-record.js
```diff
@@ -14,7 +14,7 @@ var measures = require('./../measures/measures-data.json');
 var SUBMISSION_METHOD_MAP = {
   'claims': 'claims',
   'registry': 'registry',
-  'registry/qcdr': 'qcdrOrQualifiedRegistry',
+  'registry/qcdr': 'registry',
```
kencheeto: hm, is the `registry/qcdr` key still needed?

marimiyachi: It is - this is a field in the historical benchmarks csv.

PR #24, opened 2017-04-20
Make NPM module public

marimiyachi:
https://jira.cms.gov/browse/QPPA-388

Small edits to make this module public. Should I bump the version to 1.0.0?

Reviewers: @kencheeto @gabesmed


gabesmed:
hm, maybe save that for SOFT LAUNCH????

marimiyachi:
Ok, we will use 0.0.x until all measures are published, upon which we will bump the version to 1.0.0.

PR #25, opened 2017-04-26
QPPA-516 Use qualityId as measureId

sungchoi:
### Description:
We are updating quality measurement measureIds to be equal to their qualityIds.

From JIRA Ticket:
The QPP API for quality data created a measureId that is a concatenation of qualityId and other ID fields. This field is NOT useful so we'd like to, for quality measures, remove the qualityId field and replace the existing measureId with what is now the qualityId.

This includes making follow up changes in benchmarks.

Update submissions API since historical benchmarks are keyed by measureId.

### Automated Test Plan

- [x] `npm test` should be green
- [x] `cat measures/measures-data.json | node scripts/validate-data.js measures ` should return `Valid!`
- [x] `cat benchmarks/2017.json | node scripts/validate-data.js benchmarks` should return `Valid!`

### Manual Test Plan

- [x] Code Review
- [x] `measures/measures-data.json` should be updated
- [x] `benchmarks/2017.json` should be updated


scripts/convert-qpp-to-measures.js
```diff
@@ -118,7 +118,7 @@ process.stdin.on('end', () => {
  *    emsr_id               | eMeasureId
  *    nqf_emsr_num          | nqfEMeasureId
  *    nqf_num               | nqfId
- *    qlty_id               | qualityId
+ *    qlty_id               |
```
abarciauskas-bgse: Hmm maybe this should be removed altogether - are there other fields we don't use?

marimiyachi: the measure_id row should be deleted an this should map qlty_id to measureId

scripts/convert-qpp-to-measures.js
```diff
@@ -142,69 +142,72 @@ function parseQpp(json) {
...
+      } else if (key === 'measure_type') {
+        obj.measureType = formatString(value);
+      } else if (key === 'emsr_id') {
+        obj.eMeasureId = parseId(value);
+      } else if (key === 'nqf_emsr_num') {
+        obj.nqfEMeasureId = parseId(value);
+      } else if (key === 'nqf_num') {
+        obj.nqfId = parseId(value);
+      } else if (key === 'qlty_id') {
+        obj.qualityId = parseId(value);
```
marimiyachi: This should be removed.

benchmarks/benchmarks-schema.yaml
```diff
@@ -8,7 +8,7 @@ definitions:
     type: object
     properties:
       measureId:
-        description: measureId corresponds to a measure object's measureId
+        description: measureId corresponds to a measure object's measureId. For quality measures, measureId is the same as qualityId.
```
marimiyachi: I think this line is not necessary - all references to qualityId should be removed.

scripts/convert-qpp-to-measures.js
sungchoi: Aha! That makes sense.

benchmarks/benchmarks-schema.yaml
sungchoi: Got it.

scripts/convert-qpp-to-measures.js
sungchoi: Good catch.

PR #26, opened 2017-05-11
Add transition exclusion measures

marimiyachi:
https://jira.cms.gov/browse/QPPA-603

Policy has requested that we add explicit exclusion measures, ACI_TRANS_LVOTC_1 to apply to ACI_TRANS_HIE_1 and ACI_TRANS_LVPP_1 to apply to ACI_TRANS_EP_1. 

Confirmed that schema is still valid.

Will release in Sprint 3.4

### DIFF
--- a/measures/measures-data.json
+++ b/measures/measures-data.json
@@ -86,12 +86,44 @@
     "firstPerformanceYear": 2017,
     "lastPerformanceYear": null,
     "metricType": "boolean",
+    "measureId": "ACI_TRANS_LVPP_1",
+    "title": "Low-Volume Permissible Prescriptions",
+    "description": "A MIPS eligible clinician (EC) who writes fewer than 100 permissible prescriptions during the performance period is eligible for exclusion from the required e-prescribing measure.",
+    "isRequired": false,
+    "weight": 0,
+    "measureSets": [
+      "transition"
+    ],
+    "isBonus": false,
+    "objective": "electronicPrescribing"
+  },
+  {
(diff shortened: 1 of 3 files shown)


bijujoseph:
@StevenSzeliga can you please review: There are 3 exclusion measures, but only 2 of them has TRANS peers. Why we missed one?

abarciauskas-bgse:
🤔 I think we were making an assumption before that clients would understand the dependency irregardless of the measure set - right? I _think_ the only difference is clients can pre-select all the measures for the transition measure set - but they still need to understand this dependency right? By dependent I just mean these measures are exclusionary measures such that when attested to, the corresponding entity is not responsible for submitting additional measures.

Just curious if my understanding is correct.

LGTM!

marimiyachi:
@bijujoseph I believe it's because ACI_HEI_2 does not have a transition exclusion associated with it.

marimiyachi:
@bijujoseph and @sietekk for visibility - please get back to me within the next day if you have issues with this PR :)

marimiyachi:
@abarciauskas-bgse there's been a huge volume of emails about this :) What you're saying is correct. From a technical perspective, it's not strictly necessary but it's a request from policy and not a huge deal on our end.

joshboley:
@marimiyachi will this be merged soon?  I will need these changed in our new spring for the qpp-submmission-client.

marimiyachi:
@joshboley this has been merged

PR #27, opened 2017-05-17
QPPA-594 Regenerate topped out benchmarks

marimiyachi:
https://jira.cms.gov/browse/QPPA-594

Topped out non-inverse benchmarks should be represented as [0, 0, 0, 0, 0, 0, 0, 0, 100].

Test: npm run test => Valid!

Reviewers: @sungchoi


marimiyachi:
Thanks @sungchoi  that's good to note!

PR #28, opened 2017-05-20
QPPA-569: Add claims measures fields to measures-schema.yaml

abarciauskas-bgse:
Add claims measures fields as detailed in the [tech spec](https://docs.google.com/document/d/1gaCA8bWMfJFANds9Uokxiz3TqX3rBL6bQ5tD6pz3HA4/edit#) and supplementary [data format worksheet](https://docs.google.com/spreadsheets/d/1WfoaFY_fkcKzWIO-9DaBFR_AKg0-i5ngqOidS8HrVN0/edit#gid=0).

[QPPA-569](https://jira.cms.gov/browse/QPPA-569)

Test plan:
```
cat test/examples/quality/019.json | node scripts/validate-data.js measures
cat test/examples/quality/052.json | node scripts/validate-data.js measures
cat test/examples/quality/141.json | node scripts/validate-data.js measures
cat test/examples/quality/204.json | node scripts/validate-data.js measures
cat test/examples/quality/254.json | node scripts/validate-data.js measures
cat measures/measures-data.json | node scripts/validate-data.js measures
```

should return:

```
Valid!
Valid!
Valid!
Valid!
Valid!
Valid!
```

OR

```
# Includes tests for the above scripts
npm run test
```

Work to import the measures from the origin file, [2017_Claims_IndivMeasures_SingleSource.xslx](https://docs.google.com/spreadsheets/d/1u_WA9RtIdLAIg5RLA-gDruBRpHpxuPxv1NpoWYZCsso/edit#gid=1523486830), into this repository using this schema is tracked in [QPPA-630](https://jira.cms.gov/browse/QPPA-630). The test files listed above will be reused for testing the output of the function convert origin measures data into the qpp-meaures-data format.


measures/measures-schema.yaml
```diff
@@ -110,6 +110,14 @@ definitions:
         description: Possible methods for submitting performance data for the measure.
         type: array
         items: { $ref: #/definitions/methods }
+      eligibilityOptions:
+        description: Eligibility options mirror denominator options in measure specifications. Each option comprises a set of codes used to identify eligible instances of the associated measure.
```
marimiyachi: Might want to specify that these are QCDR specifications.

measures/measures-schema.yaml
```diff
@@ -197,3 +205,55 @@ definitions:
       - thoracicSurgery
       - urology
       - vascularSurgery
+
+  eligibilityOption:
+    type: object
+    properties:
+      submissionMethods: { $ref: #/definitions/qualityCodesSubmissionMethods }
```
marimiyachi: Is there a reason to limit these?

measures/measures-schema.yaml
```diff
@@ -197,3 +205,55 @@ definitions:
...
+      sexCode:
+        enum: [M, F]
```
marimiyachi: This can be null as well right?

marimiyachi: Same for minAge and maxAge - see fields like lastPerformanceYear for how we have to handle that.

measures/measures-schema.yaml
```diff
@@ -197,3 +205,55 @@ definitions:
...
+        type: array
+        items: { $ref: #/definitions/codeObject }
+    required: [submissionMethods, optionType, qualityCodes]
+
+  codeObject:
+    type: object
+    properties:
+      code:
+        type: string
+      modifiers:  { $ref: #/definitions/arrayOfStringIdentifiers }
```
marimiyachi: Do we want to enumerate modifiers? There are only four options right?

measures/measures-schema.yaml
abarciauskas-bgse: I was thinking if it is null that key/value would be omitted entirely. Is the reason to include the `keyName: null` for consistency or another reason? It will bloat the file a lot - there will be a lot of null sexCodes, minAge and maxAge.

measures/measures-schema.yaml
abarciauskas-bgse: There's a limited set in use at the moment (more than 4) but there is I believe a huge space of possible modifiers. (see [Modifier Reference Guide](https://www.sccma-mcms.org/portals/19/assets/docs/modifier-reference-guide.pdf)) but perhaps @akegan can weigh in on if it makes sense to limit the possible set.

measures/measures-schema.yaml
abarciauskas-bgse: From discussion with @akegan I believe there are measure specifications for these code sets only for claims and registries. (see downloads from https://qpp.cms.gov/docs/QPP_quality_measure_specifications.zip)

measures/measures-schema.yaml
abarciauskas-bgse: Not sure I understand - I thought these specifications were also for claims to quality processing of claims, not sure how QCDRs factor in, could you expand?

measures/measures-schema.yaml
akegan: there are 6 or so, but I think that they could be basically anything.

measures/measures-schema.yaml
akegan: these codes that we're adding only apply to claims

abarciauskas-bgse: @akegan thanks - at the moment I've limited options to 'claims' and 'registry'. I think @marimiyachi is wondering why we're limiting it to these 2 methods (there are other methods like EHR) but my understanding is those are the only two submission methods that these code specifications apply to. I guess another way to phrase this might be only registries and claims are measuring and reporting these measures using these clinical coding standards. Is that understanding correct?

abarciauskas-bgse:
@marimiyachi re https://github.com/CMSgov/qpp-measures-data/pull/28#pullrequestreview-39575296: I think the files may be useful for testing the script which converts measures from the origin spreadsheet to our format. But if not I will remove them.

Great idea to add these scripts to `npm run test` - done so here https://github.com/CMSgov/qpp-measures-data/pull/28/files#diff-0bc9532baf4ad020545f82cf1e47d2a3

measures/measures-schema.yaml
marimiyachi: Sure! There are different types of specifications (for example, eCQM specifications versus QCDR specifications). It would be helpful to be explicit when you say "mirror denominator options in measure specifications" that you mean "QCDR measure specifications" because right now these (a) only apply to claims (b) may confuse someone if they are looking at an eCQM spec and can't find mention of "denominator options".

measures/measures-schema.yaml
marimiyachi: Right - sounds like you're correct.

measures/measures-schema.yaml
marimiyachi: Gotcha!

measures/measures-schema.yaml
```diff
@@ -197,3 +205,55 @@ definitions:
...
+      minAge:
+        type: integer
+      maxAge:
+        type: integer
+      diagnosisCodes: { $ref: #/definitions/arrayOfStringIdentifiers }
+      additionalDiagnosisCodes: { $ref: #/definitions/arrayOfStringIdentifiers }
+      procedureCodes:
+        type: array
+        items: { $ref: #/definitions/codeObject }
+    required: [submissionMethods]
```
marimiyachi: Should diagnosisCodes and procedureCodes be required?

measures/measures-schema.yaml
```diff
@@ -197,3 +205,55 @@ definitions:
...
+
+  performanceOption:
+    type: object
+    properties:
+      submissionMethods: { $ref: #/definitions/qualityCodesSubmissionMethods }
+      optionType:
+        enum: [performanceMet, performanceNotMet, performanceExclusion]
```
marimiyachi: This will likely be changing as a head's up! Waiting for final approval but performanceExclusion will likely be divided into eligablePopulationExclusion and eligiblePopulationException (field names TBD).

measures/measures-schema.yaml
marimiyachi: 👍

marimiyachi:
Cool! Asking because it seems high maintenance - up to you though!

measures/measures-schema.yaml
abarciauskas-bgse: Ok good to know - thanks!

measures/measures-schema.yaml
abarciauskas-bgse: Cool, thanks so much for that detail. I've updated the descriptions with the QCDR modifier in https://github.com/CMSgov/qpp-measures-data/pull/28/commits/a6e8e6d360cccc2addf13e70426c0c96c54088e3

measures/measures-schema.yaml
akegan: bad news, We don't have that info in the single source right now. There's no distinction in use for claims and we'd have to go through the pdfs to find the info.

marimiyachi: Yikes - which data isn't in single source? the difference between exclusion and exception?

akegan: yep, both are PD_X for us. From Manoj:
"For C2Q we have confirmed from PQMM that there is no differentiation between exclusion and exception the way we (C2Q) calculate measures. I am not sure if that’s the case for non-claim measures. In multiple meetings, as recent as on 05/18 walkthrough we have heard from the program that “exclusion” is different than “exception” and should be captured separately." 

which seems pretty ambiguous to me about whether they should be presented independently in this json. We definitely do not need them differentiated for our calculations, and they are definitely not specified explicitly in the single source excel doc.

measures/measures-schema.yaml
abarciauskas-bgse: I think one or the other is required. Do you know if there is a way to specify this?

abarciauskas-bgse: They are required in any `any of` type of relationship, e.g. procedureCodes or diagnosisCodes may present a valid eligibility option, updated the implementation here: https://github.com/CMSgov/qpp-measures-data/pull/28/commits/c26c9011f176c82cc4be7cdc8aa22b3c17d8a87a

measures/measures-schema.yaml
marimiyachi: Yeah, we've just learned that we'll have to track them as different fields. Looking at the measure 128 spec for example ([2017_Measure_128_Claims.pdf](https://github.com/CMSgov/qpp-measures-data/files/1023336/2017_Measure_128_Claims.pdf)) - the codes are different for "denominator exception" versus "denominator exclusions". Does this mean that for the implementation of this we will need to doing PDF parsing?

abarciauskas-bgse: I think at the moment, given the current version of the [single source spreadsheet](https://github.com/CMSgov/claims-to-quality-analyzer/blob/master/data/qpp_measures/2017_Claims_IndivMeasures_SingleSource.xlsx) or [here in google drive](https://docs.google.com/spreadsheets/d/1u_WA9RtIdLAIg5RLA-gDruBRpHpxuPxv1NpoWYZCsso/edit#gid=538916146) we would have to parse PDFs to differentiate betwixt these 2 options.

But I think @akegan asks for changes to be made to that spreadsheet (by the contractor? or PQMM?) based on the logic required to make that file programmatically useful.

But given pushback on changes, it might be more expedient to parse PDFs.

I'll add this comment thread as an outstanding question in the tech spec and implementation ticket.

tetraptych: Double yikes. From our conversations with PQMM, it sounds like the single source spreadsheet is pretty much finalized for 2017. I'm doubtful that separating exclusion codes and exception codes is something they could do for the current year's document.

One small ray of sunlight is that the single source document can help during QA of the PDF parsing process.

PR #30, opened 2017-05-23
[QPPA-498] Add cahps measures

abarciauskas-bgse:
QPPA-498: Add cahps measures

Closes #30 

https://jira.cms.gov/browse/QPPA-498

Description: 
Adds a scripts/add-cahps-measures.js to import CAHPS measures from an origin CSV file.

Test plan:
`npm run test`

Reviewers: @kencheeto @marimiyachi @samskeller


scripts/add-cahps-measures.js
```diff
@@ -0,0 +1,91 @@
...
+    description: '', // TBD: Will be provided by RAND,
+    nationalQualityCode: null,
+    measureType: 'patientEngagementExperience',
+    measureId: 'CAHPS_' + (idx + 1),
+    eMeasureId: null,
+    nqfEMeasureId: null,
+    nqfId: nqfIdMap[measureTitle] || null,
+    isInverse: false,
+    strata: [],
+    isHighPriority: true,
```
lucasmbrown-usds: Are all CAHPS measures high priority?

lucasmbrown-usds:
@abarciauskas-bgse  - I agree that the quite substantial commands you laid out in #3 and #4 could be combined into a single script for clarity and ease of use. It'd be nice to have a single command to generate all of the measures data from scratch. 

Up to you whether that's simple enough to be part of this PR or deserves its own ticket in the backlog. Either way, make sure to update the README.

test/measures/measures-data-spec.js
```diff
@@ -31,5 +31,40 @@ describe('measures data json', function() {
...
+
+      it('are all floats', () => testExpectedField('metricType', 'float'));
+      it('all have measureType \'patientEngagementExperience\'', () => testExpectedField('measureType', 'patientEngagementExperience'));
+      it('all have primarySteward \'Agency for Healthcare Research & Quality\'', () => testExpectedField('primarySteward', 'Agency for Healthcare Research & Quality'));
+      it('all have submissionMethods \'certifiedSurveyVendor\'', () => testExpectedField('submissionMethods', ['certifiedSurveyVendor']));
+      it('all have measureSets \'generalPracticeFamilyMedicine\'', () => testExpectedField('measureSets', ['generalPracticeFamilyMedicine']));
+      it('all have firstPerformanceYear 2017', () => testExpectedField('firstPerformanceYear', 2017));
+      it('all have category quality', () => testExpectedField('category', 'quality'));
+      it('all have isHighPriority true', () => testExpectedField('isHighPriority', true));
+      it('all have isInverse false', () => testExpectedField('isInverse', false));
```
kencheeto: would one `deepEqual` call per CAHPS measure that verifies the fields make this simpler?

```js
it('creates 12 correct CAHPS measures', function() {
  const cahpsMeasures = measuresData.filter(measure => measure.measureId.match(/CAHPS_\d+/));
  const commonCahpsProperties = {
    'measureType': 'patientEngagementExperience',
    ...
  };
  const nqfIdMap = {
    'CAHPS for MIPS SSM: Getting Timely Care, Appointments and Information': '0005',
    'CAHPS for MIPS SSM: How Well Providers Communicate': '0005',
    'CAHPS for MIPS SSM: Patient\'s Rating of Provider': '0005',
    'CAHPS for MIPS SSM: Courteous and Helpful Office Staff': '0005'
  }
  assert.equal(cahpsMeasures.length, 12);
  cahpsMeasures.forEach(cahpsMeasure => {
    assert.match(cahpsMeasure.measureTitle, /^CAHPS for MIPS/);
    // these are the same for all CAHPS measures
    assert.deepEqual(cahpsMeasure.pick(Object.keys(commonCahpsProperties)), commonCahpsProperties);
    if (nqfIdMap[cahpsMeasure.measureTitle]) {
      assert.equal(cahpsMeasure.nqfId, '0005');
    }
  });
});
```
the mocha failure message gives a nice object diff: https://mochajs.org/#diffs

also, trivia, but [passing arrow functions to mocha is discouraged](https://mochajs.org/#arrow-functions) :(

scripts/add-cahps-measures.js
```diff
@@ -0,0 +1,91 @@
+// Libraries
+const fs    = require('fs');
+const parse = require('csv-parse');
+const path  = require('path');
+
+const additionalMeasuresFilepath = '../util/additional-measures.json'
+var additionalMeasures = require(additionalMeasuresFilepath);
+
+// We want to overwrite CAHPS measures every time we run this script.
```
kencheeto: we should do this in the `parse` callback, since that's where we're inserting the measures, and it's quick to understand if these are adjacent

scripts/add-cahps-measures.js
```diff
@@ -0,0 +1,91 @@
+// Libraries
+const fs    = require('fs');
+const parse = require('csv-parse');
```
kencheeto: missing a corresponding package.json update?

scripts/add-cahps-measures.js
kencheeto: @lucasmbrown-usds Yeah, according to https://qpp.cms.gov/measures/quality:
screen shot 2017-05-23 at 5 03 49 pm

scripts/add-cahps-measures.js
```diff
@@ -0,0 +1,91 @@
...
+additionalMeasures = additionalMeasures.filter(function(measure) {
+  var re = /CAHPS_\d+/i;
+  return measure.measureId.match(re) === null;
+});
+
+// Constants
+const CAHPS_CSV_COLUMNS = [
+  'Measure Name',
+  'Measure Points',
+  'Measure Contribution'
```
kencheeto: we don't use the latter two, do we? could we remove those?

scripts/add-cahps-measures.js
```diff
@@ -0,0 +1,91 @@
...
+];
+
+// Blurgh
```
marimiyachi: Nit remove comment.

scripts/add-cahps-measures.js
```diff
@@ -0,0 +1,91 @@
...
+const defaultNqfId = '0005';
+const nqfIdMap = {
+  'CAHPS for MIPS SSM: Getting Timely Care, Appointments and Information': defaultNqfId,
+  'CAHPS for MIPS SSM: How Well Providers Communicate': defaultNqfId,
+  'CAHPS for MIPS SSM: Patient\'s Rating of Provider': defaultNqfId,
+  'CAHPS for MIPS SSM: Courteous and Helpful Office Staff': defaultNqfId
+};
+
+// Data
```
marimiyachi: Style comment nit - in general comments should explain why a code is functioning rather than what it is doing. In this case for example, the variable name is "data" so a comment marking it as data isn't needed.

scripts/add-cahps-measures.js
abarciauskas-bgse: Is it ? https://github.com/CMSgov/qpp-measures-data/blob/master/package.json#L16 I think sung used it for his benchmark work so it already exists

scripts/add-cahps-measures.js
```diff
@@ -0,0 +1,91 @@
+// Libraries
+const fs    = require('fs');
+const parse = require('csv-parse');
+const path  = require('path');
+
+const additionalMeasuresFilepath = '../util/additional-measures.json'
```
abarciauskas-bgse: Good catch! https://github.com/CMSgov/qpp-measures-data/pull/30/commits/02c989962f1d2fd1f0239f112bfdec79be19fc3b

scripts/add-cahps-measures.js
abarciauskas-bgse: +1 https://github.com/CMSgov/qpp-measures-data/pull/30/commits/bf1773e11f6353bb9e27c82006cc0fc1e3f2c481

scripts/add-cahps-measures.js
kencheeto: oh no, sorry, was looking in the wrong project 😥

test/measures/measures-data-spec.js
```diff
@@ -31,5 +31,40 @@ describe('measures data json', function() {
 
       assert.equal(intersection.size, multiPerformanceIds.size);
     });
+
+    describe('CAHPS measures', function() {
+      const cahpsMeasures = measuresData.filter(function(measure) {
+        var re = /CAHPS_\d+/i;
+        return measure.measureId.match(re) !== null;
+      });
```
abarciauskas-bgse: I think your right the function could be simpler but I thought arrow functions were discouraged? But updating the function callback to just `measure.measureId.match(/CAHPS_\d+/));`

scripts/add-cahps-measures.js
abarciauskas-bgse: I tried and failed. But I added a comment to make it clear that they are not being used and why I left them in.

test/measures/measures-data-spec.js
abarciauskas-bgse: Great improvement, updated

scripts/add-cahps-measures.js
```diff
@@ -0,0 +1,91 @@
...
+  return {
+    category: 'quality',
+    firstPerformanceYear: 2017,
+    lastPerformanceYear: null,
+    metricType: 'float',
+    title: measureTitle,
+    description: '', // TBD: Will be provided by RAND,
+    nationalQualityCode: null,
+    measureType: 'patientEngagementExperience',
+    measureId: 'CAHPS_' + (idx + 1),
```
abarciauskas-bgse: +1 updated

test/measures/measures-data-spec.js
kencheeto: arrow functions are discouraged when passing as args to mocha's `describe`/`it`/`beforeEach`/`afterEach` since `this` is lexically bound (not to mocha's test runner context, which has useful helpers like `this.timeout` to configure the timeout, etc). arrow function are great for writing code tersely in a functional style though:

```js
doSomething()
  .then(foo => foo.bar)
  .then(bar => bar.baz);
```

measures/measures-schema.yaml
```diff
@@ -15,7 +15,7 @@ definitions:
         enum: [ia, quality, aci, cost]
       metricType:
         description: Type of measurement that the measure requires in order to attest.
-        enum: [boolean, proportion, singlePerformanceRate, multiPerformanceRate, continuous]
+        enum: [boolean, proportion, singlePerformanceRate, multiPerformanceRate, continuous, float]
```
kencheeto: we might want a CAHPS-specific metricType, since there's additional information beyond the float that needs to be gathered - maybe you and @samskeller (who is working on adding the new value type) can decide on a name? 

this is a weak reference to the measurement types published at https://cmsgov.github.io/qpp-submissions-docs/schemas# and we haven't finalized the `continuous` metricType, so this doesn't need to be final

test/measures/measures-data-spec.js
```diff
@@ -31,5 +31,41 @@ describe('measures data json', function() {
 
       assert.equal(intersection.size, multiPerformanceIds.size);
     });
+
+    describe('CAHPS measures', function() {
+      const cahpsMeasures = measuresData.filter(function(measure) {
+        return measure.measureId.match(/CAHPS_\d+/);
+      });
+
+      it('creates 12 correct CAHPS measures', function() {
```
kencheeto: shoot I suggested `creates` but really should read `contains`, i think

scripts/cahps-consts.js
```diff
@@ -0,0 +1,28 @@
...
+  'CAHPS for MIPS SSM: Care Coordination': 8,
+  'CAHPS for MIPS SSM: Courteous and Helpful Office Staff': 9,
+  'CAHPS for MIPS SSM: Helping You Take Medications as Directed': 10,
+  'CAHPS for MIPS SSM: Stewardship of Patient Resources': 11,
+  'CAHPS for MIPS SSM: Between Visit Communication': 12
+};
+
+module.exports = {
+  nqfIdMap: nqfIdMap,
+  cahpsTitleToMeasureIdIndexMap: cahpsTitleToMeasureIdIndexMap
```
kencheeto: there's a nice [es6 property shorthand](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Object_initializer#New_notations_in_ECMAScript_2015) when keys match their value names:

```js
module.exports = {
  nqfIdMap,
  cahpsTitleToMeasureIdIndexMap
}
```

scripts/add-cahps-measures.js
```diff
@@ -0,0 +1,89 @@
+// Libraries
```
kencheeto: tiny nit: maybe a node shebang here, since this script is already unixy?

measures/measures-data.json
```diff
@@ -142,6 +142,294 @@
     "objective": "healthInformationExchange"
   },
   {
+    "category": "quality",
+    "firstPerformanceYear": 2017,
+    "lastPerformanceYear": null,
+    "metricType": "float",
+    "title": "CAHPS for MIPS SSM: Getting Timely Care, Appointments and Information",
+    "description": "",
```
samskeller: The Tech Spec said that RAND would provide the `description` field for these measures later on -- should we create a JIRA ticket to do that so we don't forget?

measures/measures-schema.yaml
kencheeto: `cahps` sgtm

measures/measures-data.json
abarciauskas-bgse: Good idea: https://jira.cms.gov/browse/QPPA-667

measures/measures-schema.yaml
abarciauskas-bgse: Updated!

scripts/cahps-consts.js
abarciauskas-bgse: Good call, updated

test/measures/measures-data-spec.js
abarciauskas-bgse: good catch, updated

PR #32, opened 2017-05-26
[QPPA-606] uuid and strata extraction from ecqm xml files

kencheeto:
This PR adds two scripts to help integrate necessary strata and UUID details for eCQM measures from a zip of eCQM specs for [QPPA-606](https://jira.cms.gov/browse/QPPA-606).

The new data being added for eCQMs:
* measure UUIDs are programmatically extracted from zips
  > with the exception of CMS145v5 and CMS160v5, which were extracted manually by me since their data format isn't consistent with the others
* strata descriptions and strata UUIDs are extracted from the zips
* strata names are written by me, hopefully to the effect of summarizing their descriptions
* metricType is deduced to either `singlePerformanceRate` or `multiPerformanceRate` depending on if they have one or multiple strata
* overallAlgorithm is cross-referenced from a perf rate xls file linked in the JIRA issue

Generally, the new scripts work like this:
#### `node scripts/get-strata-and-uuids-from-ecqm-zip.js <path to zip>`
1. unzips the file containing the specifications for all eCQM measures
1. parses the `CMS<measure ID>v<measure version>_SimpleXML.xml` files within to extract strata descriptions, UUIDs, and metricType
1. extracted data is written to `utils/generated-ecqm-data.json`, keyed on eMeasureId (CMS137v5)

unfortunately, strata names and overallAlgorithm are not programmatically extractable from these source files, so we need another script. This one is only used once to build a scaffold for easily manually adding missing fields by hand, but is included for reuse should we need to:
#### `node scripts/find-ecqms-with-missing-data.js <path to measures-data.json>`
1. looks in measures-data.json to find eCQM measures that are missing strata names (these are also all missing overallAlgorithm)
1. builds a json file similar to measures-data.json that only includes the missing fields (as empty strings), along with strata descriptions (since these are useful for generating the strata names)
1. the two outlier eCQMs (CMS145v5 and CMS160v5) are extracted by hand from source and hardcoded into the json
1. the json file is written to `utils/manually-added-ecqm-data.json`
1. this file needs to be edited manually to add strata names based off of descriptions and overallAlgorithm based on another source (xref'd from an xls file)

third, merge-measures-data.js has been updated to merge in all of the new generated and manual eCQM data, preserving some fields where data is already present (some eCQM strata already had names).

## Pretty messy stuff - I want to call out a few things specifically for review:

1. [strata names](https://github.com/CMSgov/qpp-measures-data/pull/32/commits/ac73a35fafb8692d1f24bf265bf19d45a6f9e0d2): I wrote these based on descriptions, need someone to read through and sanity check that they aren't misleading
1. [manually extracted eCQMs](https://github.com/CMSgov/qpp-measures-data/pull/32/commits/f501ebc4e651b9dd66d5ae1db47e08dc51bfb552): The UUIDs are correct but the strata names _and_ descriptions for 145 & 160 are handwritten based on source files from the zip. These felt a little shaky on the wording
1. our general approach for assembling measures-data.json. See changes to the merge-measures-data.js script - merging in [generated](https://github.com/CMSgov/qpp-measures-data/pull/32/commits/fc5be5b3bf1ba754ead3efbc821ee4a938766cab) and [manually added](https://github.com/CMSgov/qpp-measures-data/pull/32/commits/62195d9dd5d62c5bccb4c3a465229806dd4f75ca) data. The current merge-measures-data approach is order-dependent and each source is treated inconsistently (I'm not helping here). Relatedly, we also need to cut our dependency on the measures API which is still used to pull in the bulk of measures - if this repo is the source of truth then we should anoint a file in this repo as the measures origin. I think there's some zooming out and refactoring to do.

## Things that don't need review:

1. the 2 new scripts - feel free to take a look, but I need to refactor to get from working to maintainable
1. the measures-data.json file, this was validated already

todos:
* [x] unzip nested xmls
* [x] extract data from xmls
* [x] verify strata UUID ordering against source data
* [x] update merge-measures-data.js to integrate extracted and manually added data
* [x] manually add CMS145v5 and CMS160v5 💦 
* [x] manually add strata names 💦 
* [x] regenerate measures-data.xml
* [x] add ecqm fields to measures-schema.yaml
* [x] refactor pass


bijujoseph:
@marimiyachi , @kencheeto  - this change has no impact on QS and MA code right ?

kencheeto:
@bijujoseph eCQM measures have been updated with new strata (and corresponding changes to metricType, overallAlgorithm) - there should be no changes required for QS/MA code (no new fields), but some tests might need to be updated on your end.

util/manually-added-ecqm-data.json
```diff
@@ -0,0 +1,329 @@
...
+        "name": "receivedVaccinesOrExposure",
+        "description": "Children who have evidence showing they received recommended vaccines, had documented history of the illness, had a seropositive test result, or had an allergic reaction to the vaccine by their second birthday"
+      }
+    ]
+  },
+  {
+    "eMeasureId": "CMS123v5",
+    "strata": [
+      {
+        "name": "receivedExams",
```
samskeller: Is it worth being more specific here? Maybe `receivedFootExams` or even `receivedVisualPulseAndSensoryFootExams`?

util/manually-added-ecqm-data.json
```diff
@@ -0,0 +1,329 @@
...
+        "description": "Patients who had at least one face-to-face visit with a practitioner with prescribing authority during the Initiation Phase, and at least two follow-up visits during the Continuation and Maintenance Phase. One of the two visits during the Continuation and Maintenance Phase may be a telephone visit with a practitioner."
+      }
+    ],
+    "overallAlgorithm": "simpleAverage"
+  },
+  {
+    "eMeasureId": "CMS137v5",
+    "strata": [
+      {
+        "name": "treatmentWithin14Days",
```
samskeller: Maybe `treatmentWithin14DaysOfDiagnosis`?

util/manually-added-ecqm-data.json
```diff
@@ -0,0 +1,329 @@
...
+        "description": "Patients who initiated treatment within 14 days of the diagnosis"
+      },
+      {
+        "name": "treatmentPlusServicesWithin30",
```
samskeller: Might be worth adding "days" to the end of it? Ie `treatmentPlusServicesWithin30DaysOfVisit`?

util/manually-added-ecqm-data.json
```diff
@@ -0,0 +1,329 @@
...
+        "name": "testedForChlamydia",
+        "description": "Women with at least one chlamydia test during the measurement period"
+      }
+    ]
+  },
+  {
+    "eMeasureId": "CMS155v5",
+    "strata": [
+      {
+        "name": "BMI",
```
samskeller: Maybe `recordedHeightWeightBMI`?

util/manually-added-ecqm-data.json
```diff
@@ -0,0 +1,329 @@
...
+        "description": "Patients who had counseling for physical activity during a visit that occurs during the measurement period"
+      }
+    ],
+    "overallAlgorithm": "simpleAverage"
+  },
+  {
+    "eMeasureId": "CMS158v5",
+    "strata": [
+      {
+        "name": "hepBTested",
```
samskeller: `hepatitisBTested` just for clarity?

util/manually-added-ecqm-data.json
```diff
@@ -0,0 +1,329 @@
...
+        "name": "assessedSuicideRisk",
+        "description": "Patients with a suicide risk assessment completed during the visit in which a new diagnosis or recurrent episode was identified"
+      }
+    ]
+  },
+  {
+    "eMeasureId": "CMS166v6",
+    "strata": [
+      {
+        "name": "noImagingStudy",
```
samskeller: `noImagingStudyWithin28Days`?

util/manually-added-ecqm-data.json
```diff
@@ -0,0 +1,329 @@
...
+        "description": "Patients without an imaging study conducted on the date of the outpatient or emergency department visit or in the 28 days following the outpatient or emergency department visit"
+      }
+    ]
+  },
+  {
+    "eMeasureId": "CMS167v5",
+    "strata": [
+      {
+        "name": "dilatedMacularOrFundusExam",
```
samskeller: `dilatedMacularOrFundusExamWithin12Months`?

util/manually-added-ecqm-data.json
```diff
@@ -0,0 +1,329 @@
...
+        "description": "Patients who had a dilated macular or fundus exam performed which included documentation of the level of severity of retinopathy AND the presence or absence of macular edema during one or more office visits within 12 months"
+      }
+    ]
+  },
+  {
+    "eMeasureId": "CMS169v5",
+    "strata": [
+      {
+        "name": "alcoholAssessmentEvidence",
```
samskeller: Might be misleading to just have `alcohol` in the name since the description says "alcohol or other substance use"? Maybe `alcoholOrOtherSubstanceAssessmentEvidence`?

util/manually-added-ecqm-data.json
```diff
@@ -0,0 +1,329 @@
...
+        "name": "patientsWithReferral",
+        "description": "Number of patients with a referral, for which the referring provider received a report from the provider to whom the patient was referred"
+      }
+    ]
+  },
+  {
+    "eMeasureId": "CMS52v5",
+    "strata": [
+      {
+        "name": "within3MonthsCD4<200/mm3",
```
samskeller: Maybe prefix with `prescribed`?

util/manually-added-ecqm-data.json
```diff
@@ -0,0 +1,329 @@
...
+        "description": "Patients who were prescribed Pneumocystis jiroveci pneumonia (PCP) prophylaxis within 3 months of CD4 count below 200 cells/mm3"
+      },
+      {
+        "name": "within3MonthsCD4<500/mm3OrCD4<15%",
```
samskeller: Maybe prefix with `prescribed`?

util/manually-added-ecqm-data.json
```diff
@@ -0,0 +1,329 @@
...
+        "name": "reportedFunctionalStatus",
+        "description": "Patients with patient-reported functional status assessment results (eg, VR-12, VR-36,PROMIS-10-Global Health, PROMIS-29, HOOS) not more than 180 days prior to the primary THA procedure, and at least 60 days and not more than 180 days after THA procedure"
+      }
+    ]
+  },
+  {
+    "eMeasureId": "CMS65v6",
+    "strata": [
+      {
+        "name": "lowerFollowupBP",
```
samskeller: `lowerFollowupBloodPressure`?

util/manually-added-ecqm-data.json
```diff
@@ -0,0 +1,329 @@
...
+        "description": "Patients who have received antidepressant medications for at least 180 days (6 months) of continuous treatment during the 231-day period following the Index Prescription Start Date"
+      }
+    ],
+    "overallAlgorithm": "simpleAverage"
+  },
+  {
+    "eMeasureId": "CMS136v6",
+    "strata": [
+      {
+        "name": "hadPractitionerVisit",
```
samskeller: `hadPractitionerVisitWithin30Days`?

util/manually-added-ecqm-data.json
```diff
@@ -0,0 +1,329 @@
...
+          "denominatorExceptionUuid": "F79E2A13-6F9E-4B84-99E0-4603E99C83C3"
+        }
+      }
+    ]
+  },
+  {
+    "eMeasureId": "CMS149v5",
+    "strata": [
+      {
+        "name": "assessedCognition",
```
samskeller: `assessedCognitionOnceWithin12Months`?

util/manually-added-ecqm-data.json
```diff
@@ -0,0 +1,329 @@
...
+      }
+    ]
+  },
+  {
+    "eMeasureId": "CMS145v5",
+    "overallAlgorithm": "weightedAverage",
+    "metricType": "multiPerformanceRate",
+    "strata": [
+      {
+        "description": "Beta Blocker Therapy for LVSD",
```
samskeller: I might be very off-base about this, but it looks like from the description between this stratum and the next that the difference is just about the therapy being "for LVSD" vs not, right? Looking at the "CMS145v4.html" file in the zip, though, I also see differences in the two "Population Criteria" in the "Denominator" sections... Should those differences be reflected in these descriptions? And the general description near the top of the file mentions patients who have coronary heart disease -- should that be mentioned in these descriptions too? (Again, might be very wrong here, I still don't really understand what this all means)

kencheeto: Yeah, my descriptions were way too slim. I took a deeper look and produced these:

> Patients with at least 2 encounters with eligible professionals and left ventricular systolic dysfunction (LVEF <40%) who are prescribed beta-blocker therapy for LVSD. For these patients, beta-blocker therapy includes the following: bisoprolol, carvedilol, or sustained release metoprolol succinate.

and 

> Patients with at least 2 encounters with eligible professionals with a prior (resolved within last 3 years) myocardial infarction who are prescribed beta-blocker therapy. For patients with prior MI, beta-blocker therapy includes any agent within the beta-blocker drug class.

kencheeto:
thanks @samskeller ! updated

measures/measures-schema.yaml
```diff
@@ -82,6 +82,9 @@ definitions:
       eMeasureId:
         description: Identifier for Electronic Clinical Quality Measures (ECQM).
         type: ['null', string]
+      eMeasureUuid:
```
marimiyachi: Are these required fields now for quality measures?

scripts/find-ecqms-with-missing-data.js
```diff
@@ -0,0 +1,119 @@
...
+    // add stub for overallAlgorithm if needed
+    if (measure.strata.length > 1 && !measure.overallAlgorithm) {
+      scaffold.overallAlgorithm = '';
+    }
+    return scaffold;
+  });
+
+const missingJson = outlierEcqms.concat(ecqmsWithMissingData);
+const sortedMissingJson = _.sortBy(missingJson, ['eMeasureId']);
+fs.writeFileSync(path.join(__dirname, '../util/manually-added-ecqm-data.json'), JSON.stringify(sortedMissingJson, null, 2));
```
marimiyachi: Maybe write to standard out instead? Scripts that use standard in/out are little more flexible.

scripts/merge-measures-data.js
```diff
@@ -47,9 +47,70 @@ function mergeQpp(qppJson) {
...
+      ecqmInfo.strata.forEach((newStratum, ecqmIndex) => {
+        if (oldStrata[ecqmIndex]) {
+          qppJson[index].strata[ecqmIndex].eMeasureUuids = newStratum.eMeasureUuids;
+          qppJson[index].strata[ecqmIndex].description = newStratum.description;
+        } else {
+          qppJson[index].strata[ecqmIndex] = newStratum;
+        }
+      });
+      if (oldStrata.length > 1) {
+        // ASSUMPTION: strata already available are in the same order as the ones extracted from the xml files
```
marimiyachi: 😳

measures/measures-schema.yaml
marimiyachi: Looks like no, but they are also not nullable.

scripts/find-ecqms-with-missing-data.js
```diff
@@ -0,0 +1,119 @@
...
+  }
+];
+
+const measuresJson = JSON.parse(fs.readFileSync(measuresDataPath, 'utf8'));
+const ecqmsWithMissingData = measuresJson
+  .filter(measure => measure.category === 'quality')
+  .filter(measure => measure.strata.some(stratum => !stratum.name))
+  .map(measure => {
+    const strata = measure.strata
+      .filter(stratum => !stratum.name)
```
marimiyachi: Why do you need to filter again here? Are there some measures that only one of the stratum's have names?

marimiyachi:
Also GREAT JOB ON THIS. Amazing comments - really easy to follow along what is actually a super complicated process.

scripts/find-ecqms-with-missing-data.js
kencheeto: Yeah, some eCQMs have one named stratum, so need to filter those out.

scripts/find-ecqms-with-missing-data.js
kencheeto: agree! i will do that in the refactor pass.

scripts/find-ecqms-with-missing-data.js
kencheeto: (I'll add a comment to explain this)

scripts/merge-measures-data.js
```diff
@@ -47,9 +47,70 @@ function mergeQpp(qppJson) {
...
       }
+
     }
   });
 
   console.error('did not find measure details for the following', measuresNotFound);
+
+  // separate passes for ecqm data to keep things simple
+  // load extracted ecqm data
+  const generatedEcqmStrataJson = JSON.parse(fs.readFileSync(path.join(__dirname, '../util/generated-ecqm-data.json'), 'utf8'));
```
marimiyachi: Does it make sense to handle this inline with the other performance rate data file processing (line 30)? Right now the `measuresNotFound` error is compiling measures that are likely represented here.

marimiyachi: That way the override logic for individual fields could be more explicit?

measures/measures-schema.yaml
kencheeto: Ahhh, I misunderstood how fields are optional unless defined explicitly - will make this a string-only type and leave it optional.

scripts/merge-measures-data.js
kencheeto: Yeah, I started that way, but wanted to keep them separate at first since we could reason about the new additions separately without also considering the original logic in merge-measures-data. For example, I ran into an issue with the guard clause exiting early when no perf rate info was found - I didn't want to change the resulting logic to consider when ecqm data was found but not perf rate data (and vice versa).

I'm thinking that the refactor can explicitly enumerate the different sources of data and the fields pulled from them.

scripts/merge-measures-data.js
kencheeto: The ordering is derived from the spec, so as long as we are generally always pulling things in the correct order from the spec we should be okay. Still, yeah, kinda 😬

scripts/find-ecqms-with-missing-data.js
marimiyachi: Hm....It's a bit of a strange circular dependency going on here where we use `measures-data.json` to generate `manually-added-ecqm-data.json` and then use `manually-added-ecqm-data.json` to generate `measures-data.json`. Are you sure that this step is necessary? Presumably what is in `measures-data.json` pre this PR about eCQMs is invalid and after this PR it should be corrected (i.e. all stratums should have names).

scripts/find-ecqms-with-missing-data.js
marimiyachi: Cool - maybe add a TODO?

scripts/merge-measures-data.js
marimiyachi: Hm....ok I hear ya. The logging for `measuresNotFound` should be adjusted at the very least.

scripts/find-ecqms-with-missing-data.js
kencheeto: this script isn't meant to be run more than once unless new ecqms are added - maybe I shouldn't include it at all? it just generates an empty scaffold that can easily be edited by hand.

marimiyachi:
Last comment - in general with the strata names I'd say lean on the side of being as concise as possible. It's meant to be a differentiator between the other strata, not to provide all description.

scripts/merge-measures-data.js
kencheeto: I didn't find anything to add to `measuresNotFound` - I've already added logging for potential issues when adding ecqm data, so I moved the `measuresNotFound` flushing to the bottom to make it clear that it happens at the end.

scripts/find-ecqms-with-missing-data.js
kencheeto: fixed!

scripts/find-ecqms-with-missing-data.js
kencheeto: i added clarifying instructions to this script, and it's not referenced in the readme, so I believe it shouldn't be confusing anymore: https://github.com/CMSgov/qpp-measures-data/pull/32/commits/758025dd5f326935926a5918ca6c96f5356f9234

scripts/merge-measures-data.js
kencheeto: i added some comments to merge-measures-data to make the ecqm additions clearer: https://github.com/CMSgov/qpp-measures-data/pull/32/commits/b3fb1243c5ac53e040e9b2de5b320db2b60e3705

kencheeto:
thanks @marimiyachi 🙏 ! I think I've addressed your comments (shortened strata names in https://github.com/CMSgov/qpp-measures-data/pull/32/commits/b7bab5fc4d2830647b1ec35203c57ed3af0bb146), this should be good to merge now! I wrote a followup issue [QPPA-687](https://jira.cms.gov/browse/QPPA-687) for refactoring the scripts. As-is, this PR incrementally worsens the complexity of the measures-data.json build process, but should be well documented enough to allow debugging/maintenance.

QPPA-687 will take a step back and includes suggestions for refactoring the build process.

PR #38, opened 2017-06-20
QPPA-783: Shortening strata names as they need to be less than 20 characters

samskeller:
Strata names need to be 20 characters or shorter to fit in the column in the database, so we need to shorten the names of strata that are too long. Also I've changed the schema file to enforce the 20-character limit on the name field so that a JSON file will be invalid if the names are too long. Lastly, bumping version of package up to 0.0.34

https://jira.cms.gov/browse/QPPA-783

As a separate thought, should we put in a pre-commit hook that checks the JSON file against the schema to make sure it's valid?

### DIFF
--- a/measures/measures-data.json
+++ b/measures/measures-data.json
@@ -3140,7 +3140,7 @@
           "denominatorExclusionUuid": "672f9875-55d9-4fb3-b56f-85205699103f",
           "denominatorExceptionUuid": "3a57a106-7404-476e-b026-23282a95d75f"
         },
-        "name": "evidenceOfSubstanceUseBetweenDiagnosisAndTreatment"
+        "name": "substanceUse"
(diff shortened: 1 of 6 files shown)


lucasmbrown-usds:
@samskeller would be great to do your idea of "pre-commit hook that checks the JSON file against the schema" -- can we create a backlog ticket for your product owner to prioritize? or, if it's a one liner, go for it?

samskeller:
@lucasmbrown-usds yeah I actually noticed that this checking-against-schema is happening in our automated tests, but there was no pre-commit hook for running the automated tests. I've added a PR for ensuring that the automated tests pass before someone commits something to `master`: https://github.com/CMSgov/qpp-measures-data/pull/39

PR #39, opened 2017-06-22
Adding the pre-commit npm package so we can ensure tests pass before commits to master

samskeller:
We need a way of ensuring that the automated tests pass in this
repository before new code is committed to master -- to do this, there
is now a script in tools/ that will run the automated tests if something
is committed to master and this is triggered with the pre-commit npm
package which is now apart of this repo

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,7 +10,8 @@
   "scripts": {
     "test": "node_modules/.bin/_mocha --colors",
     "build:measures": "scripts/build-measures",
-    "build:benchmarks": "scripts/build-benchmarks"
+    "build:benchmarks": "scripts/build-benchmarks",
+    "pre-commit": "tools/pre-commit.sh"
   },
   "license": "MIT",
   "devDependencies": {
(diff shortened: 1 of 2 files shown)


gabesmed:
Does this mean we need to merge to master locally, vs. using github?

samskeller:
@gabesmed hmm that's a good point, I assumed people were squash merging locally because that's what we do in other repos, but I'm now realizing that that requirement isn't documented here. I feel like it's worth it to have people merge locally so that we can have the tests run with each merge to master, like we do in `qpp-submissions-api`, but lemme know what you think -- I can update the CONTRIBUTING.md as part of this PR if you think that's the right way to go. There's not a way to disallow merging via the GitHub interface, right? We just have to hope that people do it locally?

gabesmed:
yeah we've had this github vs. local merging convo before. maybe can be discussed during portal time. in the meantime tho definitely no harm in having the pre-commit hook so that local merges do run the tests

PR #40, opened 2017-06-22
[QPPA-789] adds eMeasureUuid for CMS145v5 and CMS160v5

kencheeto:
https://jira.cms.gov/browse/QPPA-789

* includes `eMeasureUuid` for CMS145v5 and CMS160v5, which were manually handled as their eCQM xml files were not processable like the others
* it was error-prone to regenerate measures-data manually, so I collected steps to generate measures-data.{json,xml} into one script: `generate-measures-data` - this simplifies the docs significantly, though we still want to refactor the component scripts
* moves documentation for generating measures-data into [CONTRIBUTING.md](CONTRIBUTING.md) where it's clear that consumers of the repo won't need to read those steps, only contributors.


CONTRIBUTING.md
```diff
@@ -12,6 +12,45 @@ npm install
 
 Make changes on a feature branch, then open a pull request. Make sure CI passes on your branch, and you include any relevant new tests.
 
+### Generating measures-data.json
+To regenerate the `measures-data.json` and `measures-data.xml` files, which contain additional metadata and conform to
+the measures schema, do the following:
+
+```
+npm start
```
marimiyachi: Can we use a different verb here? start doesn't feel correct since the default behavior is to start a server.

CONTRIBUTING.md
```diff
@@ -12,6 +12,45 @@ npm install
...
+```
+
+### Generating benchmarks data
+To regenerate benchmarks data from historical data use the `scripts/parse-benchmarks-data.js` script
+like so:
+```
+cat data/historical-benchmarks/2015.csv | node scripts/parse-benchmarks-data.js 2015 2017
```
marimiyachi: Make this a scripts command too?

marimiyachi:
🚢

PR #41, opened 2017-06-23
QPPBSR-1367 - WI benchmark data

bijujoseph:
JIRA ticket : https://jira.cms.gov/browse/QPPBSR-1367
@marimiyachi , @samskeller - The WI Benchmark data supplied by @StevenSzeliga  is added in this file. 
"DM-Composite" , that measure either we will have to add it to measure-data json or remove it with DM-7 or DM-2 (diabetes composites). Anyway Steven is working with Policy on that.. I will raise a different PR later on for that.

### DIFF
--- a/benchmarks/2017.json
+++ b/benchmarks/2017.json
@@ -221,6 +221,23 @@
     ]
   },
   {
+    "measureId": "110",
+    "benchmarkYear": 2015,
+    "performanceYear": 2017,
+    "submissionMethod": "cmsWebInterface",
+    "deciles": [
+      0,
+      30.00,
+      40.00,
+      50.00,
+      60.00,
+      70.00,
+      80.00,
+      90.00,
+      100
+    ]
+  },
+  {
     "measureId": "111",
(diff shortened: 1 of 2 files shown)


samskeller:
@bijujoseph I'll let others with more context respond to the actual benchmark changes, but I thought I'd note that the version number was just bumped to 0.0.35 in this PR which was just merged to master: https://github.com/CMSgov/qpp-measures-data/pull/42

So perhaps you want 0.0.36?

marimiyachi:
@bijujoseph - this file is actually [generated](https://github.com/CMSgov/qpp-measures-data/blob/master/scripts/build-benchmarks) by a script, so instead of editing it directly there are two options:
1) Add these benchmarks to the [CSV file](https://github.com/CMSgov/qpp-measures-data/blob/master/data/historical-benchmarks/2015.csv) for parsing
2) Add these in a different file such as `addtional_benchmarks.json` and then concatenate that file as a part of the build benchmarks process.

PR #44, opened 2017-06-26
[QPPA-743] Ignore empty denom exceptions/exclusions from eCQM zip

kencheeto:
https://jira.cms.gov/browse/QPPA-743

The `CMSXvY_Simple.xml` files included UUIDs for denominator exception and exclusion fields that were marked as 'None'. I've refactored `get-strata-and-uuids-from-ecqm-zip.js` to parse the `CMSXvY.xml` files instead, which only provide the UUIDs in the [reference PDF](https://www.cms.gov/Regulations-and-Guidance/Legislation/EHRIncentivePrograms/Downloads/eCQM_QRDA_EC.pdf
). I've also updated 145v5 and 160v5 since those were manually pulled, removing the UUIDs for denominator fields that are blank.

I have verified that all of the changes to measures-data in this pr are correct when comparing to the reference PDF, so I only need review for the script changes 🙏

### DIFF
--- a/scripts/get-strata-and-uuids-from-ecqm-zip.js
+++ b/scripts/get-strata-and-uuids-from-ecqm-zip.js
@@ -87,9 +87,25 @@ function extractStrata(measure) {
     strata[index].eMeasureUuids = {
       initialPopulationUuid: ids.find(item => item.$.type === 'initialPopulation').$.uuid,
       denominatorUuid: ids.find(item => item.$.type === 'denominator').$.uuid,
-      numeratorUuid: ids.find(item => item.$.type === 'numerator').$.uuid,
-      denominatorExclusionUuid: ids.find(item => item.$.type === 'denominatorExclusions').$.uuid,
-      denominatorExceptionUuid: ids.find(item => item.$.type === 'denominatorExceptions').$.uuid,
+      numeratorUuid: ids.find(item => item.$.type === 'numerator').$.uuid
+    };
+
+    const denominatorException = ids
+      .find(item => (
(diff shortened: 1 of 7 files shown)


scripts/get-strata-and-uuids-from-ecqm-zip.js
```diff
@@ -62,12 +62,14 @@ return strata name, description, and uuids like so
...
   // parse out strata descriptions from numerator text
-  const description = details.numeratorDescription[0];
   // descriptions are like "Numerator 1: Patients who initiated treatment within 14 days of the diagnosis\nNumerator 2: Patients who initiated treatment and who had two or more additional services with an AOD diagnosis within 30 days of the initiation visit"
-  let strataDescriptions = _.compact(description.split('\n'))
+  const description = measure.subjectOf
+    .find(item => item.measureAttribute[0].code[0].$.code === 'NUMER')
+    .measureAttribute[0].value[0].$.value;
+
+  let strataDescriptions = _.compact(description.split(/\n|\r|&#xA;/))
     // if multiple strata, they're enumerated as 'Numerator x: '
```
marimiyachi: Why is this split changing?

kencheeto: I switched from parsing CMS1v2_Simple.xml files to CMS1v2.xml files, and the latter uses `&#xA;` (html entity for line feed character) to delimit some but not all strata descriptions :\

kencheeto:
@marimiyachi no changes to build-measures needed! i had some concerns that data sources were getting mixed up (registry pdfs vs ecqm zip) but checked out all the changes in this pull request :+1:

PR #51, opened 2017-07-12
ACI relationship with new field subcategory & substitute

bijujoseph:
The core idea is to add minimal possible meta-data in measures and update the build script to enrich the existing measures. 

The requirement from our side are: 
  Given a measure, I need to know
   1.  what kind of measure it is (example "pre-attestation" or "registry performance" etc
   2.  which other measures can substitute the given measure. 

The approach here uses a new field `subCategory` and `substitutes`. If you prefer reusing `subCategoryId` instead of the new field `subCategory` do let me know.

PR #52, opened 2017-07-12
aci relationship - measureSets update

bijujoseph:
The core idea is to add minimal possible meta-data in measures and update the build script to enrich the existing measures. 

The requirement from our side are: 
  Given a measure, I need to know
   1.  what kind of measure it is (example "pre-attestation" or "registry performance" etc
   2.  which other measures can substitute the given measure. 

The approach here uses a new field  `substitutes` and existing `measureSets`


bijujoseph:
Closing in favor of #51

PR #59, opened 2017-08-02
Feature/qppqs 442 deciles for topped out measures

rockykitamura:
Benchmark deciles for topped out measures with only decile 10 populated in the source CSV were inconsistent with deciles populated for other measures.

Also removed benchmarks where deciles are all 0's from `additional-benchmarks-2017.json`.  Specifically, `046`, `370`, and `438`, these measures do not have benchmarks.


.gitignore
```diff
@@ -1,5 +1,6 @@
 node_modules
 .DS_Store
+.idea
```
kencheeto: I mentioned this to @bijujoseph too, but might be more convenient for you to add this to your global gitignore so you don't need to add it to every repo.

rockykitamura: Thanks.  I'll remove it here and add it to my global gitignore.

PR #62, opened 2017-08-28
[QPPA-695] import csv measures

abarciauskas-bgse:
Adds scripts and documentation to convert a QCDR CSV file to QPP measures.

[QPPA-695](https://jira.cms.gov/browse/QPPA-695)

Detail:
* Adds `import-qcdr-measures.js` which is a command line script for reading a QCDR CSV file and writing the result to stdout.
* Does **not** add any data to `measures-data.json`. As noted in the config file there are a few outstanding questions about how the data should be imported that should be resolved before actually adding measures data to `measures-data.json|xml`. Most notably there are multiple rows per measureId at this time.

Test plan:
* Automated tests
* Script examples in `import-qcdr-measures.js`:

```
cat util/measures/20170825-PIMMS-non-mips_measure_specifications.csv | node ./scripts/measures/import-qcdr-measures.js

cat util/measures/20170825-PIMMS-non-mips_measure_specifications.csv | node ./scripts/measures/import-qcdr-measures.js | node scripts/validate-data.js measures
```


CONTRIBUTING.md
```diff
@@ -41,6 +41,27 @@ At this time `util/measures/additional-measures.json` includes:
...
 Similarly, `util/benchmarks/additional-benchmarks-2017.json` contains benchmark data that is necessary for scoring but not included in the historical CSV file. Any additional benchmarks should be added to this file, followed by a re-generation of benchmarks data.
 
+### Importing measures from CSV file
+
+The `scripts/measures/import-csv.js` script handles importing new measures from a CSV and converting them to the qpp-measures-data measure schema. A config file is required in order to correctly map columns in the spreadsheet map to qpp-measures-data measures fields. A config file is a yaml file with the following structure:
+
+```yaml
+sourced_fields:
+  measureId: 0
+  isHighPriority:
```
marimiyachi: This abstraction is a little off to me. It requires having pretty deep knowledge of how the `import-csv` script will use these fields (for example, it's not super clear what Y/N mean), which for me indicates that they should live together rather than separately. 

The scenario that is coming up right now (with potentially combining rows) also makes me think that the logic we need surpasses what a config file can offer. Basically if it were a simple process of mapping one column to one field, I'd support this approach, but I'm not convinced about putting logic in a config file.

I'm also not sure if we can reasonably expect to reuse this either - the tradeoff here is reusability versus readability/ease to support custom logic. Looking at how we parsed the benchmarks file, for example, there is often a lot of processing that we need to do that can't be determined by a config file. Another example - what if the data file contains multiple metricTypes, but is not specified by a column, so we need to use other logic to determine what it is for each row. I think there are a number of cases that could occur like this.

measures/measures-schema.yaml
```diff
@@ -129,6 +130,12 @@ definitions:
         description: Performance options mirror numerator options in QCDR measure specifications. Each option comprises a set of codes used to identify instances of performance met, performance not met, performance exclusion or performance exception.
         type: array
         items: { $ref: #/definitions/performanceOption }
+      isRiskAdjusted:
+        type: boolean
+        default: false
+      vendorId:
+        description: Submissions of the measure are restricted to the organization associated with vendorId.
+        type: string
     required: [nationalQualityStrategyDomain, measureType, eMeasureId, nqfEMeasureId, nqfId, isHighPriority, isInverse, strata, primarySteward, measureSets]
```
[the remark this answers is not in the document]
abarciauskas-bgse: Wouldn't that imply we need to add it to all existing quality measures?

CONTRIBUTING.md
abarciauskas-bgse: I don't disagree that most files will require custom logic, however I don't think we lose anything by writing a script which is reusable for importing csv files since we would have to implement this field mapping logic in either case. The tradeoff as I see it is we either implement:
1. a reusable script + an additional shorter script for custom logic vs
2. 1 longer script which contains all custom logic

I don't know what the future holds for requirements for measures data but this script could have been used for importing CAHPS data.

marimiyachi: @kencheeto wdyt?

samskeller: I'm no kencheeto, and I think I lack context about this issue that y'all have, but my 2 cents: I think on the whole I generally agree with @marimiyachi -- the data that we're receiving seems to be so convoluted and custom that I'm not sure that we'll be able to fit all of our parsing requirements for CSVs into a config file. While writing custom CSV parsing scripts for each CSV that we receive seems cumbersome (and I think that's the alternative here...?), I think it would be a little clearer and easier to maintain than a config abstraction that requires complicated changes to work for new CSVs. 

That being said, I don't think there's a ton of risk in trying out this abstraction and seeing if it's useful for other CSVs? 

I'm about a 4/10 on my opinion here

kencheeto: I don't think there's enough evidence for supporting csv import configured via config file, considering the cost to learn and maintain a new abstraction. For example, we would be bound to support it for external developers, which I'm wary of. I think specifying the format for a staging file and enforcing that they are additive only provides enough guidance for developers to make contributions without being prescriptive.

I do think there's appeal in making a script reusable, and I wonder if using a library like [csv](https://github.com/wdavidw/node-csv) would help with some of the boilerplate of dealing with csv files:

* selecting headers from the first line, or providing your own via callback (with the first line as input)
* process each row as an object via ['transform' function](http://csv.adaltas.com/transform/)

That function would easily be able to hardcode values, set defaults, or run arbitrary js to bookkeep and handle duplicate rows.

> I don't think there's a ton of risk in trying out this abstraction and seeing if it's useful for other CSVs?

I think once we use it I don't think it'll ever be a priority to un-use it.

samskeller: > I think once we use it I don't think it'll ever be a priority to un-use it.
Yeah, good point -- stale code is certainly a risk

bijujoseph:
Just tagging @rockykitamura as he from our side often works on enhancing measure-data.
cc: @sietekk

abarciauskas-bgse:
Fixed test writing stderr in https://github.com/CMSgov/qpp-measures-data/pull/62/commits/4afb5e4f76b10dd0ba43a045ceae867980b5a1aa

PR #64, opened 2017-08-28
QPPA-607: Add AHR benchmark data

marimiyachi:
https://jira.cms.gov/browse/QPPA-607

Manually adding a benchmark for measure 438.

Reviewers: @abarciauskas-bgse

### DIFF
--- a/benchmarks/2017.json
+++ b/benchmarks/2017.json
@@ -3891,5 +3891,22 @@
       90,
       100
     ]
+  },
+  {
+    "measureId": "458",
+    "benchmarkYear": 2015,
+    "performanceYear": 2017,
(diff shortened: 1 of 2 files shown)


benchmarks/2017.json
```diff
@@ -3891,5 +3891,22 @@
       90,
       100
     ]
+  },
+  {
+    "measureId": "458",
+    "benchmarkYear": 2015,
+    "performanceYear": 2017,
+    "submissionMethod": "cmsWebInterface",
```
marimiyachi: @bijujoseph or @StevenSzeliga can you confirm that this is the only submission method for measure 438 All-Hospital Readmission?

PR #66, opened 2017-08-29
Remove ng api dependence

abarciauskas-bgse:
Removes dependence on NG API from `build-measures`

**Detail**
* First step in implementing qpp-measures-data refactor as detailed in the [tech spec](https://docs.google.com/document/d/1FVNAEdaGKyE1z7no5Qjs6PwAu7_d46UBjlGPdQbWFD0/edit#heading=h.yf7c5gh4zq3d)
* Archives `convert-qpp-to-measures` script which implements transformations on NG API measures. Archives instead of deletes so README can still reference how the original source of truth was transformed.
* Removes the invocation of other one-time use scripts from `build-measures`. These scripts only make sense when `build-measures` can first download the original measures from qpp.cms.gov/api and build on top of them.
    * These scripts and related data files will be moved in a subsequent step.

**Test plan**
* `./build-measures` works and `measures/measures-data.json` remains unchanged.


bijujoseph:
@abarciauskas-bgse can you please share the google doc with us ? esp @rockykitamura as he often works on enhancing measure-data from our side. Thank you.

PR #71, opened 2017-09-15
Feature/qppqs 499 groups for cpc+ submissions

rockykitamura:
- Adding `cpcPlusGroup` attribute to specific quality measures to support CPC+
- Adding measure specification links to measure PDF's by submission method


scripts/measures/enrich-measures-data.js
```diff
@@ -0,0 +1,53 @@
...
+    .filter(measure => measure.category === 'quality')
+    .forEach(measure => {
+      Object.keys(cpcPlusGroups).forEach((groupId) => {
+        var match = cpcPlusGroups[groupId].find((id) => id === measure.eMeasureId);
+        if (match !== undefined) {
+          measure.cpcPlusGroup = groupId;
+        }
+      });
+    });
+};
```
joshboley: Why do these functions need to be in this new file when the same code is in `merge-measures-data.js`?

rockykitamura: Actually, I think `merge-measures-data.js` is deprecated.  I will verify and move it into the `archive` folder.

rockykitamura:
@joshboley I suppose we could copy out the original version of the `measures-data.json` file and always operate on the original file to generate the enriched version of the `measures-data.json`.  I think the intent is to keep all the functions in `enrich-measures-data.js` idempotent so this is probably a matter of preference.  I'll let @bijujoseph and @kencheeto weigh in on that decision.

test/scripts/measures/import-qcdr-measures-spec.js
```diff
@@ -12,7 +12,7 @@ const expectedMeasures = require('./fixtures/expected-measures.json');
 
 // Function which executes script and converts output to a JS object.
 const runTest = function(file) {
-  const cmd = 'cat ' + file + ' | node ./scripts/measures/import-qcdr-measures.js';
+  const cmd = 'cat ' + file + ' | node ./scripts/measures/archived/scripts/import-qcdr-measures.js';
```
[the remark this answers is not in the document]
rockykitamura: Yes, when I moved the script to archive, rather than throw the test away, I opted to keep the test around.

scripts/measures/enrich-measures-data.js
abarciauskas-bgse: @rockykitamura could we remove the new functions from merge-measures-data so there's no confusion about what script should own these functions?

abarciauskas-bgse:
@bijujoseph @rockykitamura yes, the idea is that build-measures should be an idempotent script (and subsequently enrich-measures).

@rockykitamura could you add a test plan to assert that `measures/measures-data.json` doesn't change if you run build-measures twice? This can be an local test.

I would also ask @marimiyachi to weigh in on the change to measures-schema.

scripts/measures/enrich-measures-data.js
rockykitamura: No problem.  I've removed the functions from the merge-measures-data script.

measures/measures-schema.yaml
```diff
@@ -51,6 +51,9 @@ definitions:
         description: If true, attesting to the measure will qualify the provider for an ACI CEHRT bonus.
         type: boolean
         default: false
+      measureSpecification:
```
marimiyachi: Is this required? It should be a feature of the measures schema I think?

scripts/measures/enrich-measures-data.js
```diff
@@ -0,0 +1,74 @@
+const fs = require('fs');
+const path = require('path');
+const parse = require('csv-parse/lib/sync');
+
+const aciRelations = require('../../util/measures/aci-measure-relations.json');
+const cpcPlusGroups = require('../../util/measures/cpc+-measure-groups.json');
+
+const qpp = fs.readFileSync(path.join(__dirname, '../../staging/measures-data.json'), 'utf8');
```
marimiyachi: What is the staging data file from?

scripts/measures/enrich-measures-data.js
```diff
@@ -0,0 +1,74 @@
...
+ */
+function enrichACIMeasures(measures) {
+  // add extra ACI metadata to ACI measure
+  measures
+    .filter(measure => measure.category === 'aci')
+    .forEach(measure => {
+      // find the relation and populate reporting category and substitutions
+      var aciRelation = aciRelations[measure.measureId];
+      if (aciRelation) {
+        measure.reportingCategory = aciRelation.reportingCategory;
```
marimiyachi: Do these fields need to be added to the schema?

scripts/measures/enrich-measures-data.js
```diff
@@ -0,0 +1,74 @@
...
+
+function enrichAddMeasuresSpecification(measures) {
+  var csv = parse(fs.readFileSync(path.join(__dirname, '../../util/measures/measurePDF-Specification.csv'), 'utf8'));
+  var mappedLinks = csv.reduce(function(acc, [submissionMethod, measureId, link]) {
+    acc[measureId] = acc[measureId] || {};
+    acc[measureId][submissionMethod] = link;
+    return acc;
+  }, {});
+  var measureData = measures.map(function(measure) {
+    measure['measureSpecification'] = mappedLinks[measure.measureId];
```
marimiyachi: Convention is to use dot notation.

test/scripts/measures/import-qcdr-measures-spec.js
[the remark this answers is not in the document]
marimiyachi: This script shouldn't be archived.

scripts/measures/enrich-measures-data.js
rockykitamura: The `reportingCategory` and `substitutes` attributes have already been added to the `aciMeasures` schema in a previous commit https://github.com/CMSgov/qpp-measures-data/blob/65864866f828c349ef41e55f9b42e69d457492f9/measures/measures-schema.yaml#L60

The change here related to ACI is moving the ACI enriching function to a single script with other measure enriching functions.

scripts/measures/enrich-measures-data.js
rockykitamura: The staging data file is the version of `measures-data.json` currently on the `master` branch.  https://github.com/CMSgov/qpp-measures-data/blob/65864866f828c349ef41e55f9b42e69d457492f9/measures/measures-data.json#L1

We've opted to place this original copy of `measures-data.json` in `staging` and then apply the enriching functions to the original copy to generate the version in `measures/measures-data.json`.  Let me know if you'd like this to be changed @marimiyachi.

cc: @bijujoseph

test/scripts/measures/import-qcdr-measures-spec.js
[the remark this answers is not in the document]
abarciauskas-bgse: Actually @marimiyachi the tech spec states these scripts should be archived 
> one-time custom scripts will be stored in the script archive of the repository
So if we want a different policy, let's discuss it apart from this PR.

scripts/measures/enrich-measures-data.js
marimiyachi: Gotcha. Thanks!

test/scripts/measures/import-qcdr-measures-spec.js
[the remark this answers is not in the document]
marimiyachi: 👍

PR #73, opened 2017-10-16
QPPA-1222 adding "proposed" to unfinalized ACI exclusions

ivanang:


abarciauskas-bgse:
Nice ! Can you also run `npm run build:measures`  so that `./measures/measures-data.json` and `./measures/measures-data.xml` are updated?

PR #75, opened 2017-10-17
Fix inverse measures

philipdickinson:
This PR corrects the `isInverse` attribute for measures based on the measure definition PDFs.

The script used to surface which measures should be inverse is based on the performance rate script.

---
This change is [Reviewable](https://reviewable.io/reviews/cmsgov/qpp-measures-data/75)


scripts/measures/get-inverse-measures-from-pdfs.js
```diff
@@ -0,0 +1,59 @@
+// running this will print out whether each measure is an inverse measure
```
akegan: can you update the readme to include instructions for running this script in instructions for how to generate the whole pdf

scripts/measures/get-inverse-measures-from-pdfs.js
```diff
@@ -0,0 +1,59 @@
...
+// claims pdfs can be downloaded from https://qpp.cms.gov/docs/QPP_quality_measure_specifications.zip
+// unzip and give the path to the Claims-Quality-Measures
```
akegan: This should be both claims and registry quality measures, right? same for line below

scripts/measures/get-inverse-measures-from-pdfs.js
```diff
@@ -0,0 +1,59 @@
...
+}
+
+function getIsInverse(path, cb) {
+  pdfToText.pdfToText(path, function(err, data) {
+    if (err) {
+      cb(false);
+      return;
+    }
+
+    let found = data.match(/inverse measure/i); // ignore case
```
akegan: did you do a comparison between inverse measure and INVERSE MEASURE to see if anything is picked up? I didn't see any pdfs with lower case on a spot check, but am curious to see if there are any that are lower case.

scripts/measures/get-inverse-measures-from-pdfs.js
```diff
@@ -0,0 +1,59 @@
...
+const groupedFiles = fs.readdirSync(folderPath).reduce(function(arr, current) {
+  let qualityId = current.split('_')[2];
+  arr[qualityId] ? arr[qualityId].push(current) : arr[qualityId] = [current];
+  return arr;
+}, {});
+
+const qualityIds = Object.keys(groupedFiles);
+qualityIds.forEach(function(qualityId, i) {
+  getIsInverse(p.join(folderPath, groupedFiles[qualityId][0]), function(isInverse) {
+    console.log(qualityId + ' : ' + isInverse);
```
akegan: Is this writing into the measures data file directly? would it be possible to update it to automatically update to make it easier to keep it up to date?

philipdickinson: I think that directly manipulating the measure definition file goes against the approach taken in the rest of the repo. There are updates to the file that can be applied in `scripts/measures/enrich-measures-data.js` but it didn't really feel right to add behaviour here as it would be dependent on resources from outside the repository (the PDFs).

scripts/measures/get-inverse-measures-from-pdfs.js
philipdickinson: I just ran it with lower case for you and the result was the same.

scripts/measures/get-inverse-measures-from-pdfs.js
philipdickinson: Sorry, what do you mean?

akegan: When starting at the top of the file I thought that this could be added to steps to generate the measures-data spec. Looks like those instructions don't exist anywhere though.  @marimiyachi or @jiahuang  do you have instructions anywhere for how to recreate measures-data that we could include this step in? I want to make sure that the inverse measures are generated correctly if the data has to be updated for some reason.

scripts/measures/get-inverse-measures-from-pdfs.js
akegan: @marimiyachi or @jiahuang will have a better final answer, but manually updating the file to include these inverse measure results doesn't seem like a sustainable solution to me.

scripts/measures/get-inverse-measures-from-pdfs.js
philipdickinson: I follow. I can change the script to output the results to JSON (and keep the resulting file under source control) and then add a step to `scripts/measures/enrich-measures-data.js` which would use that file. Let me do that now, should be quick 👍

scripts/measures/get-inverse-measures-from-pdfs.js
philipdickinson: I understand now I think - see my other comment above 👍

marimiyachi: Yeah - this repo is in a really weird in between state right now as we are half way through a refactor. Preferred approach would be to add this as a function in `enrich-measures-data` and programmatically updated the fields. One approach that we've used in the past is the download/gitignore the zip (see [here](https://github.com/CMSgov/qpp-measures-data/blob/68742bc66ac383f5d155a682dc8426202d35ba91/scripts/measures/build-measures)) - that's probably the most straightforward approach - you could also write the results of this into a "staging" file - that might be more work than it's worth.

philipdickinson:
Apologies for my confusion on that initial submission - updated now. The script to parse the PDFs will write the output to a file (which is under source control) and there's an extra step in `scripts/measures/enrich-measures-data.js` to use that file to populate the measure definitions correctly.

philipdickinson:
Thanks for the reminder @marimiyachi - just added the XML

PR #76, opened 2017-10-18
Feature/qppqs 628 dm composite measures fix

rockykitamura:
Change DM Composite measure id to "dmComposite"

https://jira.cms.gov/browse/QPPQS-628

### DIFF
--- a/benchmarks/2017.json
+++ b/benchmarks/2017.json
@@ -3723,7 +3723,7 @@
     ]
   },
   {
-    "measureId": "DM-Composite",
+    "measureId": "dmComposite",
     "benchmarkYear": 2015,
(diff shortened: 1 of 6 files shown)


kencheeto:
We use the measureId as a db & validation key, so I'm checking to see if there are measurements already in prod that have submitted for this measure.

bijujoseph:
@kencheeto By "prod" are you referring to "qpp.cms.gov" ? this is a WI measure and ONLY that system is supposed to submit it. So, even if someone accidentally submitted it anywhere then it is wrong.

kencheeto:
@bijujoseph agreed, I'd just like to check first if we expect to invalidate any data; folks would start to see errors

PR #80, opened 2017-11-01
bump eslint, eslint --fix

kencheeto:
eslint-config-nava 2.x adds a new rule ([no-var](https://eslint.org/docs/rules/no-var)). I've bumped the version here and ran `eslint --fix test util scripts index.js` to automatically fix new violations, and manually fixed three instances of var -> const/let.

I should have separated out the manual fixes from the automatic ones, happy to do that if anyone thinks it's important.


kalvinwang:
@kencheeto just curious, why didn't the linter catch those three instances?

kencheeto:
@kalvinwang the linter did catch those instances, but couldn't automatically fix them - there was, for example, a structure like this:

```js
if (...) {
  var foo = 'hello';
}
```

and we can't safely change the var to a let or const since the `var` is being hoisted and eslint doesn't know where you want the hoisted let/const to go (if changed in-place, the outer block scope would lose visibility of `foo`)

kencheeto:
i'm not going to bump the version, as this is functionally equivalent - it can get rolled into the next version bump.

PR #85, opened 2017-11-08
[QPPA-1310] Set metricType to multiplePerformanceRate on some QCDR measures

kalvinwang:
This PR sets `metricType` to `multiPerformanceRate` for QCDR measures with `Number of performance rates to be included in the XML` greater than 1, AND (proportional=Y, continuous=N, ratio=N).

In order to include strata names and descriptions for multiPerformanceRate measures, we've manually modified the QCDR csv in these ways for each of the 27 affected measures (aka QCDR measures where the metricType is multiPerformanceRate):

- For column N, `Indicate an Overall Performance Rate if more than 1 performance rate is to be submitted`, we modify the values to strictly be one of: 
  - the number of the rate that is the overall rate (e.g. `3`) instead of `Rate 3` or `Performance Rate 3`, or 
  - `weighted average` or `sum numerators` (case-insensitive)
- For rows where column N is filled out, in column E, `Measure Description`, we modify the values to strictly be in the format:
  - `*summary* Rate 1: description Rate 2: description Rate 3: description`, instead of e.g. 
  - `*summary* A) description B) description C) description *more summary*`

We plan to ask policy to make the above modifications prior to future updates to the QCDR csv.

Testing: checked results manually and found 27 measures change from `singlePerformanceRate` to `multiPerformanceRate` which appears to be accurate. Added a unit test, tests pass.

(This is a followup to #81, and includes a generated file that should have been added as part of that PR, `measures-data-with-qcdrs.json`)

Jira: https://jira.cms.gov/browse/QPPA-1310

- [x] Include multiple strata name/descriptions and overallalgorithm field
- [x] Include other strata data once policy responds
- [x] Move strata names to a separate file

Reviewer: @kencheeto @marimiyachi


kalvinwang:
This PR should be reviewable now! Just missing some of the strata data pending questions for policy.

kalvinwang:
In the process of adding new strata data (thanks @ivanang!), heads up @marimiyachi @kencheeto that I'm going to be removing the extra column I added and using the existing Measures Description field instead (with some manual formatting)

kalvinwang:
@marimiyachi @kencheeto all strata data carefully cleaned up and added!

scripts/measures/import-qcdr-measures.js
```diff
@@ -92,7 +87,9 @@ const config = {
  * @param  {object}           config  object defining how to build a new measure from this csv file, including mapping of measure fields to column indices
  * @return {array}            Returns an array of measures objects
  *
- * We trim all data sourced from CSVs because people sometimes unintentionally include spaces or linebreaks
+ * Notes:
+ * 1. The terms 'strata' and 'performance rate' are used interchangeably
```
marimiyachi: Maybe we should be a little more specific and say that we are referring to the performance rate strata as opposed to the reporting rate strata?

scripts/measures/import-qcdr-measures.js
```diff
@@ -120,15 +117,86 @@ const convertCsvToMeasures = function(records, config) {
...
+        const measureDescription = _.trim(record[4]);
+
+        // Measure description column contains performance rate description
+        // Split '*summary* Rate 1: text Rate 2: text' into [text, text]
+        const strata = _.split(measureDescription, /\s*[Rr]ate [0-9]+:\s*/);
+        // Drop anything before 'Rate 1' (usually a description of the measure)
+        strata.shift();
+
+        // TODO(kalvin): move strata names to a separate file
+        const STRATA_NAMES = {
```
marimiyachi: Small batch measure names :)

scripts/measures/import-qcdr-measures.js
```diff
@@ -120,15 +117,86 @@ const convertCsvToMeasures = function(records, config) {
...
+          'ICLOPS17': ['withfollowup', 'withoutfollowup']
+        };
+
+        newMeasure['strata'] = [];
+        _.each(strata, function(stratum, index) {
+          strataName = STRATA_NAMES[measureId][index];
+          // i + 1 because Rates in the csv are numbered starting from 1
+          if (_.lowerCase(strataName) === 'overall' &&
+            index + 1 !== nthPerformanceRate) {
+            throw TypeError('"Overall" strata for ' + measureId + ' in QCDR ' +
```
marimiyachi: Nice check!

scripts/measures/import-qcdr-measures.js
marimiyachi: Maybe have a comment indicating that the order of these arrays in important?

measures/measures-schema.yaml
```diff
@@ -142,7 +142,7 @@ definitions:
       cpcPlusGroup:
         description: CPC+ group which the measure belongs
         type: string
-    required: [nationalQualityStrategyDomain, measureType, eMeasureId, nqfEMeasureId, nqfId, isHighPriority, isInverse, strata, primarySteward, measureSets, isRegistryMeasure]
+    required: [nationalQualityStrategyDomain, measureType, eMeasureId, nqfEMeasureId, nqfId, isHighPriority, isInverse, primarySteward, measureSets, isRegistryMeasure]
```
marimiyachi: This might be overkill - but an alternative to removing this validation would be to create another measure type so we can validate that strata and overallAlgorithm are present when the metricType is multiPerformanceRate - similar to [here](https://github.com/CMSgov/qpp-measures-data/pull/85/files#diff-cc1568c0c7b370d047afdf355f584710R37).

scripts/measures/import-qcdr-measures.js
kalvinwang: Is this clearer or just more confusing? Please feel free to suggest an improved version :)

`The terms [performance rate] 'strata' and 'performance rates' are used interchangeably. `

scripts/measures/import-qcdr-measures.js
kalvinwang: good idea k

measures/measures-schema.yaml
```diff
@@ -142,7 +142,18 @@ definitions:
       cpcPlusGroup:
         description: CPC+ group which the measure belongs
         type: string
-    required: [nationalQualityStrategyDomain, measureType, eMeasureId, nqfEMeasureId, nqfId, isHighPriority, isInverse, strata, primarySteward, measureSets, isRegistryMeasure]
+    oneOf: [
```
marimiyachi: 👏

measures/measures-schema.yaml
kalvinwang: (done!)

PR #87, opened 2017-11-09
[QPPA-458] Add a merge script that merges json files based on specified keys and…

lamroger:
… order of files

I added a script called `scripts/benchmarks/merge-benchmark-data.js` and it builds an object with each measurement being key'd by a composite key of `measureId, benchmarkYear, performanceYear, and submissionMethod`. 

It iterates through an array of files, where later benchmarks overwrites previously defined benchmarks. Right now only changes in deciles are considered differences.

If this looks like the way we want to approach, I will write tests and integrate with the `build-benchmark` file.


lamroger:
Also want to note the jq method in `build-benchmark` only appends, not overwrites because it doesnt have the notion of differentiating keys. You probably could code something with jq but it'd be hard to understand and maintain.

scripts/benchmarks/merge-benchmark-data.js
```diff
@@ -0,0 +1,43 @@
...
+  'benchmarkYear',
+  'performanceYear',
+  'submissionMethod'
+];
+
+const getBenchmarkKey = (benchmark) => {
+  let benchmarkKey = '';
+  UNIQUE_COLUMN_CONSTRAINT.forEach((keyName) => {
+    benchmarkKey += `${benchmark[keyName]}|`;
+  });
```
kencheeto: a comment explaining what this generates would be helpful!

scripts/benchmarks/merge-benchmark-data.js
```diff
@@ -0,0 +1,43 @@
...
+
+  return benchmarkKey;
+};
+
+const mergeBenchmarkLayers = (benchmarkLayers) => {
+  const mergedBenchmarks = new Map();
+
+  benchmarkLayers.forEach((benchmarkLayer) => {
+    const benchmarkFile = require(benchmarkLayer);
```
kencheeto: let's avoid requiring inside function calls and put requires at the top of the file; while we're technically allowed to lazily require, it's good practice to follow es6 module rules for import/export, as we'll eventually switch to that module format.

scripts/benchmarks/merge-benchmark-data.js
```diff
@@ -0,0 +1,43 @@
...
+    benchmarkFile.forEach((benchmark) => {
+      mergedBenchmarks.set(getBenchmarkKey(benchmark), benchmark);
+    });
+  });
+
+  const formattedBenchmarks = [];
+  for (const value of mergedBenchmarks.values()) {
+    formattedBenchmarks.push(value);
+  }
```
kencheeto: is this equivalent to:
```js
const formattedBenchmarks = [...mergedBenchmarks.values()];
```
?

scripts/benchmarks/parse-benchmarks-data.js
```diff
@@ -15,7 +15,8 @@ var BENCHMARK_CSV_COLUMNS = [
   'decile8',
   'decile9',
   'decile10',
-  'isToppedOut'
+  'isToppedOut',
+  ''
```
kencheeto: is this so you have a trailing comma in the generated csv? might be worth having a comment here

scripts/benchmarks/merge-benchmark-data.js
```diff
@@ -0,0 +1,43 @@
...
+  return benchmarkKey;
+};
+
+const mergeBenchmarkLayers = (benchmarkLayers) => {
+  const mergedBenchmarks = new Map();
+
+  benchmarkLayers.forEach((benchmarkLayer) => {
+    const benchmarkFile = require(benchmarkLayer);
+    benchmarkFile.forEach((benchmark) => {
+      mergedBenchmarks.set(getBenchmarkKey(benchmark), benchmark);
```
kencheeto: is full replacement of previous benchmarks OK? or are there any fields included in previous benchmark files that we want to keep despite new info in subsequent layers?

marimiyachi: Full replacement seems like the best way forward - for maintainability and given the way we receive updates is via updated CSVs (i.e. it would be easy to miss an update if we were only looking in certain columns).

scripts/benchmarks/merge-benchmark-data.js
```diff
@@ -0,0 +1,43 @@
+const BENCHMARK_LAYERS = [
+  '../../staging/benchmarks/benchmarks_base_2017.json',
+  '../../staging/benchmarks/benchmarks_updates_20171109.json',
+  '../../staging/benchmarks/benchmarks_readmission_20171109.json'
```
marimiyachi: Why is the readmission one separate? Could you document the process for adding new benchmarks in the future? It looks like you ran parse-benchmarks-data to generate these files?

marimiyachi: I wonder if we could make this step a little less hardcoded - so iterate though the files in the benchmarks directory and merge based no date stamps?

scripts/benchmarks/merge-benchmark-data.js
lamroger: yea, it seemed sloppy to me too. How does `    const benchmarkFile = JSON.parse(fs.readFileSync(path.join(__dirname, benchmarkLayer), 'utf8'));
` look?

lamroger: borrowed it from the `enrich-measures-data.js` file

scripts/benchmarks/merge-benchmark-data.js
lamroger: Cool. Double checking, what im currently doing is a full replacement? I still need to overall key on a set of keys.

scripts/benchmarks/merge-benchmark-data.js
lamroger: Yeah, I'll add more info in the CONTRIBUTING file on how to add new benchmarks.

I'm thinking every time we need to add or modify a benchmark, we should add a new json file for auditing and regeneration incase there's a bug. Ya, I used parse-benchmarks-data and I'll document how to use that too.

We should make the ordering explicit because ordering matters when we merge. We could make it less hard coded by appending 001-<name>, 002-<name> and have ordering by file name or include in the code the ordering as I'm doing it now.

WDYT?

scripts/benchmarks/parse-benchmarks-data.js
lamroger: Ya it is, I'll add a comment.

scripts/benchmarks/merge-benchmark-data.js
lamroger: yep. nice!

scripts/benchmarks/merge-benchmark-data.js
kencheeto: hm, works for me!

scripts/benchmarks/merge-benchmark-data.js
kencheeto: Yeah this is a 'full replacement' because the entire benchmark value is getting replaced (not merging via Object.assign).

scripts/benchmarks/merge-benchmark-data.js
kencheeto: I think because we're doing full replacement, there isn't 'modifying' a benchmark, just replacing with more recent copies, which means ordering will matter. Explicit ordering via dates in filenames makes sense to me (which could get parsed in this script)

test/benchmarks/merge-benchmark-data-spec.js
```diff
@@ -0,0 +1,155 @@
...
+          14.14,
+          9.09,
+          3.33,
+          0
+        ]
+      }];
+      assert.deepEqual(mergeBenchmarkData.mergeBenchmarkLayers(layers), resultingBenchmarks);
+    });
+
+    it('joins a third independant layer', function() {
```
kencheeto: `s/independant/independent`

lamroger: i need a spell check addon that isnt super annoying

PR #92, opened 2017-11-15
[QPPA-458][Bug Fix] Include benchmarks that were excluded during conversion

lamroger:


staging/benchmarks/json/001_benchmarks_updates.json
```diff
@@ -3162,6 +3434,40 @@
     ]
   },
   {
+    "measureId": "54",
```
lamroger: Do we have a standard format for measureIds? Should I pad numbers with leading zeros?

What happens in this PR is that if its not recognized in the measures file, it uses the measureId in the benchmark CSV which doesn't have leading 0s

scripts/benchmarks/format-benchmark-record.js
```diff
@@ -127,6 +133,34 @@ const formatDecileGenerator = function(record) {
   };
 };
 
+// Looks in measures-data.json for an existing measureIds and
+// walks through the original, capitalized + removed spaces,
+// and capitalized + underscored spaces versions.
+// If found, returns the measureId from the measures-data.json file.
+// If none are found, return the padded number or non-spaced version
+const formatMeasureId = (measureId) => {
+  const measureIdRemovedSpaces = measureId.replace(/\s+/g, '');
```
marimiyachi: Are there any cases with both spaces AND underscores?

scripts/benchmarks/format-benchmark-record.js
```diff
@@ -127,6 +133,34 @@ const formatDecileGenerator = function(record) {
...
+
+  for (const formattedMeasureID of measureIdCandidates) {
+    const measure = MEASURE_ID_TO_MEASURE_MAP[formattedMeasureID];
+    if (measure) return measure.measureId;
+  };
+
+  // If all digits, pad with zeros up to the thousands
+  // else, return a nonspaced version
+  if (measureId.match(/^\d+$/)) {
+    return ('0000' + measureId).slice(-4);
```
marimiyachi: What's the reason for padding in this case?

lamroger: It's the format numeric-only benchmarks are using right now. There isn't a set standard but I wanted to conform to something we have already.

lamroger: They're actually padded to the hundredth place in `benchmarks/2017.json`. Made the change and added a test.

marimiyachi: 👍

scripts/benchmarks/format-benchmark-record.js
lamroger: Ya, it's rare but it can happen. I'll write some regex to fuzzy match

lamroger: Rare as it someone would have to generate it and I havent seen it before

test/scripts/benchmarks/format-benchmark-record-spec.js
```diff
@@ -1,20 +1,43 @@
...
 // Utils
-const formatBenchmarkRecord = require('./../../../scripts/benchmarks/format-benchmark-record');
+const { formatBenchmarkRecord, formatMeasureId } = require('./../../../scripts/benchmarks/format-benchmark-record');
 
 const options = {
   benchmarkYear: 2016,
   performanceYear: 2018
 };
 
+describe('formatMeasureId', function() {
```
marimiyachi: nice tests :)

test/scripts/benchmarks/parse-benchmark-data-spec.js
```diff
@@ -0,0 +1,21 @@
...
+
+// Function which executes script and reads in output file to a JS object.
+const runTest = function(benchmarksCsv) {
+  const cmd = 'cat ' + testCsv + ' | ' + 'node ./scripts/benchmarks/parse-benchmarks-data.js 2015 2017';
+  const benchmark = execSync(cmd).toString();
+  return JSON.parse(benchmark);
+};
+
+describe('parse benchmark data', function() {
+  it('should create new measures and ignore duplicate measureIds', () => {
```
kencheeto: if we throw an error when a fuzzy matched measure id can't be found, could we reduce this test case down to checking for duplicate measureIds (using a smaller test csv, checking for specific measure ids in output)? this is a hard test case to maintain; if it breaks i'm not sure how I'd debug what went wrong.

test/scripts/benchmarks/parse-benchmark-data-spec.js
```diff
@@ -0,0 +1,21 @@
...
+    const measures = runTest(testCsv);
```
kencheeto: would `benchmarks` be a better variable name?

PR #99, opened 2017-12-14
[QPPA-1447] Update metricTypes for qcdr measures

abarciauskas-bgse:
Updates qcdr measures with `isRegistryMeasure === true` to use new metric types `registrySinglePerformanceRate` and `registryMultiPerformanceRate`

Steps:
* Write and run script to update staging file
* Updated `measures/measures-schema.yaml`
* Run `./scripts/measures/build-measures`

Changes:
* Updated staging file for qcdr measures
* Updated measures-data.json and measures-data.xml
* Updated measures-schema.yaml with new metric types

```javascript
const fs = require('fs');

qcdrMeasuresDataPath = './staging/measures-data-with-qcdrs.json';
qcdrMeasuresString = fs.readFileSync(qcdrMeasuresDataPath, 'utf-8');
qcdrMeasures = JSON.parse(qcdrMeasuresString);

updatedMeasures = [];
qcdrMeasures.forEach((measure) => {
  if (measure.isRegistryMeasure === true) {
    if (measure.metricType === 'singlePerformanceRate') {
      measure.metricType = 'registrySinglePerformanceRate';
    } else if (measure.metricType === 'multiPerformanceRate') {
      measure.metricType = 'registryMultiPerformanceRate'
    }
  }
  updatedMeasures.push(measure);
});

updatedMeasuresJson = JSON.stringify(updatedMeasures, null, 2);
fs.writeFileSync(qcdrMeasuresDataPath, updatedMeasuresJson);
```

Testing:
* measures data is valid
```
⇒  cat measures/measures-data.json | node scripts/validate-data.js measures
Valid!
```
* By inspection, number of measures updated is equal to number of measures where `isRegistryMeasure: true` ("isRegistryMeasure": true - 610 matches, isRegistryMeasure and nonProportion metricType - 78 matches, isRegistryMeasure and regisrySinglePerformanceRate or registryMultiplePerformanceRate metricTypes - 532 measures)

PR #100, opened 2017-12-19
[QPPA-1451] Update QCDR Measures (from MedAmerica)

kalvinwang:
This PR updates a few QCDR measures from nonProportion to singlePerformanceRate.

* All MBSAQIP measures were updated to only Proportional
* E-CPR11 measure was updated to only Ratio

- [x] check if these measures exist on dev/imp/val/prod; if so the db will need to be cleared
Ran these sql queries in each env; they exist in dev but not in the other three environments:
`select measure_id from non_proportion_measurements where measure_id LIKE 'ECPR11'`
`select measure_id from non_proportion_measurements where measure_id LIKE 'MBSAQIP%'`

Jira: https://jira.cms.gov/browse/QPPA-1451

Testing: updated tests, passes all tests

Thanks to @saquino0827 and @abarciauskas-bgse for help investigating!

Reviewer: @abarciauskas-bgse


scripts/measures/enrich-measures-data.js
```diff
@@ -100,7 +100,7 @@ function enrichStratifications(measures) {
     .filter(measure => measure.category === QUALITY_CATEGORY)
     .forEach(measure => {
       const stratification = stratifications.find(stratum => stratum.eMeasureId === measure.eMeasureId);
-      if (stratification) {
+      if (stratification && stratification.mapping) {
```
kalvinwang: @saquino0827 can you take a quick look at this-- I got an error when running this script and this seems like the right fix; some of the strata are being removed from measures-data.json but that looks like a consequence of two of these measures switching from nonproportional to singleperformancerate? Anyway I figure you're more familiar with the intent. Thanks!

scripts/measures/enrich-measures-data.js
```diff
@@ -100,7 +100,7 @@ function enrichStratifications(measures) {
     .filter(measure => measure.category === QUALITY_CATEGORY)
     .forEach(measure => {
       const stratification = stratifications.find(stratum => stratum.eMeasureId === measure.eMeasureId);
-      if (stratification) {
+      if (stratification && stratification.strataMaps) {
```
kalvinwang: @saquino0827 can you take a quick look at this-- I got an error when running this script and this seems like the right fix; some of the strata are being removed from measures-data.json but that looks like a consequence of two of these measures switching from nonproportional to singleperformancerate? Anyway I figure you're more familiar with the intent. Thanks!

kalvinwang: chatted with @saquino0827 over hipchat and resolved this-- the underlying issue was that I hadn't pulled from latest master, which had a bugfix for `stratification.strataMaps` (previously incorrectly `stratification.mapping`) below, which restored all the missing strata.

scripts/measures/import-qcdr-measures.js
```diff
@@ -93,7 +93,7 @@ const addMultiPerformanceRateDetails = function(newMeasure, record, qcdrStrataNa
   const strataNames = fs.readFileSync(path.join(__dirname, qcdrStrataNamesDataPath), 'utf8');
   const qcdrStrataNames = JSON.parse(strataNames);
 
-  newMeasure['metricType'] = 'multiPerformanceRate';
+  newMeasure['metricType'] = 'registryMultiPerformanceRate';
```
kalvinwang: this is related to/necessary for a previous pr: https://github.com/CMSgov/qpp-measures-data/pull/99

PR #104, opened 2018-01-03
QPPA-1409: Update benchmarks from HCQAR file

samskeller:
This PR adds in changes to the benchmarks from an HCQAR file (see [QPPA-1409](https://jira.cms.gov/browse/QPPA-1409)). The only modifications made to the file was changing the column names to match what `scripts/benchmarks/parse-benchmarks-data.js` expects and to export the data as a CSV (since it came to us as an excel file). I then placed that file into `staging/benchmarks/csv/` and ran:
```
cat staging/benchmarks/csv/003_benchmarks_hcqar_updates.csv | node scripts/benchmarks/parse-benchmarks-data.js 2015 2017
```
to generate the JSON. Then I ran `npm run build:benchmarks`, which updated the `2017.json` benchmarks file.

I also noticed a small mistake in the `CONTRIBUTING.md` guide, so I've fixed that in this PR.

Lastly, bumping the package version number from `1.0.1` to `1.0.2`.

Here is a screenshot of the two relevant lines from the original CSV to compare against the changes to `2017.json` generated:
screen shot 2018-01-03 at 1 00 22 pm
The changes in `2017.json` look sensible in light of this data -- seems like the `parse-benchmarks-data.js` script added a first decile of 100 for ACEP21 and 0 for ACEP24 in addition to what was in the file (though it didn't add a decile at the end -- I'm assuming that's expected)


samskeller:
cc @rockykitamura @bijujoseph this will mean we'll need a new version of the scoring engine once this is merged so we can start using these measures in the Submissions API

staging/benchmarks/json/003_benchmarks_hcqar_updates.json
```diff
@@ -0,0 +1,4388 @@
+[
```
abarciauskas-bgse: Do we need this file for any reason?

abarciauskas-bgse: Now that I understand better what's going on I think what should be changed is storing benchark csv's in the `/staging/benchmarks/csv` directory. I understand storing csv's in that directory was a convention from before this PR so does not need to be changed as part of this PR. However, wanted to mention that data in `/staging/*` is intended to be valid measures or benchmark json (or a valid subset of fields which are joinable on `measureId`) according the respective schemas and this data meant to be reserved for data that is used during the `build:measures` or `build:benchmarks` scripts.

The place for source data such as `staging/benchmarks/csv/003_benchmarks_hcqar_updates.csv` is actually `/util/benchmarks`. Which perhaps itself should be changed from `util` to `source_data` or just `data`

Let me know what you think.

samskeller: @abarciauskas-bgse so are you saying that _everything_ in `staging/benchmarks/csv/` should be moved over to `util/benchmarks/`? Or just the `003` file? And either way, the `util/benchmarks/` directory is kinda confusing... right now, we have a CSV, a JSON file, and a JS script in there. We could make a `csv` directory in there to put the stuff from `staging/benchmarks/csv/` into as well as the `2017.csv` in there, but not sure what to do with the `additional-benchmarks-2017.json` file...? Thoughts?

bijujoseph:
@samskeller thanks for the update! we will update on our side once these changes get released.

staging/benchmarks/json/003_benchmarks_hcqar_updates.json
abarciauskas-bgse: I agree `/util/benchmarks/` is confusing, but think we should leave it as-is until we are prepared to implement a more encompassing solution.

PR #106, opened 2018-02-23
QPPA-1588: Updating metricType of measureId 458 (ACR) to be 'cost'

samskeller:
In QPPA-1588, we'll be adding a new database table to store data for cost-related measures, like ACR (even though ACR is technically a quality measure) and future cost measures. This PR changes the `metricType` of ACR, which is measure 458, to be `cost` and updates our schemas to expect this as a valid `metricType`.

Note -- I'm wondering if it'll be confusing if we have a `metricType` of `cost` if we're eventually also going to have a `category` of `cost`. Should we name the `metricType` something else to save that confusion? ` But what...?

### DIFF
--- a/measures/measures-data.json
+++ b/measures/measures-data.json
@@ -2872,7 +2872,7 @@
     "category": "quality",
     "firstPerformanceYear": 2017,
     "lastPerformanceYear": null,
-    "metricType": "cost",
+    "metricType": "costScore",
     "title": "All-cause Hospital Readmission",
     "description": "The 30-day All-Cause Hospital Readmission measure is a risk-standardized readmission rate for beneficiaries age 65 or older who were hospitalized at a short-stay acute care hospital and experienced an unplanned readmission for any cause to an acute care hospital within 30 days of discharge.",
     "nationalQualityStrategyDomain": "CCC",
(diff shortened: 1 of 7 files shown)


samskeller:
@kencheeto definitely not confident, but that's been the guiding theory so far -- that cost measures are going to look similar in that they will have a top-level number score with a bunch of "metadata" that aren't actually used by the scoring engine but are useful for feedback. That being said, it might be smart to name the metricType something more general in case all cost-category-specific measures don't look like that. Any thoughts on a better name...? `scoreAndMetadata`? `scoreWithJSON`? `costScoreAndMetadata`? We should probably set the database table name in https://github.com/CMSgov/qpp-submissions-api/pull/887 to match, too -- ie `score_and_metadata_measurements` or something

marimiyachi:
What about instead of relying on the measureId in this PR https://github.com/CMSgov/qpp-submissions-api/pull/887/files, we create more specific metric types (so in this case something like `costScore` and have multiple metric types map to the cost measurements table?

marimiyachi:
Another possibility would be calling this metric type `administrativeClaims` - talked to Steven and he said that phrasing makes sense in the context of ACR. `costScore` sounds more descriptive to me though.

marimiyachi:
Yeah that sounds reasonable - probably the best approach depends on how similar or different future cost measures are from one another - so we can go with that latter approach and adjust if needed later!

PR #107, opened 2018-03-01
QPPA-1132: Changing metricType for measure 321 from singlePerformanceRate to cahps

samskeller:
Measure 321 is the aggregate measure for CAHPS and currently has a metric type of 'singlePerformanceRate'. The aggregate value for CAHPS score will be a percentage (just like the individual CAHPS measurements) so we should use the CAHPS metric type for 321 as well. This PR changes that metricType for measure 321.

Also updating the version number to be `1.0.5` so it can be published and consumed by downstream services.

Reviewers: @kencheeto @lamroger

### DIFF
--- a/measures/measures-data.json
+++ b/measures/measures-data.json
@@ -3670,7 +3670,7 @@
     "category": "quality",
     "firstPerformanceYear": 2017,
     "lastPerformanceYear": null,
-    "metricType": "singlePerformanceRate",
+    "metricType": "cahps",
     "title": "CAHPS for MIPS Clinician/Group Survey",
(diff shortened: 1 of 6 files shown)


kencheeto:
cc @rockykitamura @bijujoseph - the metric type is changing (from single perf rate to cahps) for this CAHPS measure, in case that's relevant for the scoring engine.

bijujoseph:
thanks for the update @samskeller

bijujoseph:
@samskeller - l discussed with Steven he seems to have a few questions. 
In my view point "metricType" defines the kind of measurement (or the value of you can expect in a measurement). I am not sure the shape/fields we expect in `measurement.vaule` for CHAPS-1 is same as 321. If 321 is not aligning with singlePerformanceRate structure, probably we should create a new one and not rebrand it as "chaps".

samskeller:
Talked in hipchat with @bijujoseph and Steven -- we're basically just making use of the `cahps` metricType so we can store the float value between 0 and 100 necessary for measureId 321 and the scoring engine will ignore the `mask`, `reliability`, and `isBelowMinimum` values for any measurements with a measureId of 321

PR #108, opened 2018-03-14
Add behaviour for including claims-to-quality-related attributes to measures JSON

philipdickinson:
This is for https://jira.cms.gov/browse/QPPA-630 and https://jira.cms.gov/browse/QPPA-633 (via https://jira.cms.gov/browse/QPPCTOQ-280).

This PR contains:
* Python scripts for generating a JSON representation of relevant measures data from measures spreadsheet. (These are adapted from those used in C2Q)
* Support for running the above inside a container, for those who don't have the Python or the relevant libraries.
* Initial data, generated using the above.
* Changes to `enrich-measures-data` to add a step which integrates the above data into the final JSON file.
* Updated `measures-data.{json|xml}` to incorporate the above changes.
* A brief readme describing how to use the scripts to parse the spreadsheet.

Outstanding questions:
* ~Do I also need to update `/measures/measures-schema.yaml`?~

EDIT: I suppose the CI system has answered my outstanding question there! Let me update the schema - feel free to wait for that before reviewing.


philipdickinson:
Hi @marimiyachi @kencheeto - I'm not sure how to proceed with fixing the schema. The output of validation is `Invalid: data[141] should match exactly one schema in oneOf`. That is measure 093, but doesn't indicate which level oneOf is the offending one, or whether it matches 0 or 2+ schemata. Looking at the JSON for that measure I don't see any obvious problems. Is there a way to get more detailed info out of the validation process?

kencheeto:
@philipdickinson I tried ajv's verbose mode as well as inspecting the error object that was returned, but nothing more useful than the current test failure output. Might be fastest to isolate measure 093 and start cutting out the schema & measure json until you identify where it's broken?

philipdickinson:
Just tried stepping through the validation in the debugger for just that measure. It's really not an easy thing to debug since the actual validation function is generated according to the schema. I guess I'll just need to trial-and-error it with the different schema attributes.

philipdickinson:
Alright, turned out to be simple in the end. Moving forward I'd be in favour of validation that provides more useful output - happy to help out if you ever decide to do a little refactor.

claims-related/README.md
```diff
@@ -0,0 +1,20 @@
...
+Run the script from the container:
+```bash
+docker-compose run data-analysis python /home/jovyan/work/scripts/single_source_to_json.py
+```
+The docker-compose file in this directory mounts the correct directories for this operation using the default paths. Other paths can be specified with the following arguments:
+
+* `--single_source_filepath`
+    * Path to the spreadsheet containing the claims-related data, e.g. `/claims-related/data/2017_Claims_IndivMeasures_SingleSourceVersion2_11082017_andrea_edit`
+* `--measures_data_filepath`
+    * Path to the measures data JSON, e.g. `/measures/measuers-data.json`
```
akegan: typo

claims-related/scripts/single_source_to_json.py
```diff
@@ -0,0 +1,117 @@
...
+    )
+
+    parser.add_argument(
+        '-i', '--single_source_filepath',
+        help='Single Source Excel file input filename',
+        required=False,
+        default=DEFAULT_SINGLE_SOURCE_EXCEL_PATH,
+        type=str)
+
+    parser.add_argument(
```
akegan: is this still necessary?

philipdickinson: All three arguments can be used, so not much reason to remove any imo

philipdickinson: Oh wait, yeah the input JSON arg isn't used in this version. I'll remove it.

claims-related/README.md
```diff
@@ -0,0 +1,20 @@
+# Summary
+This directory contains resources for parsing the single source spreadsheet to produce a JSON file containing measure eligibility and performance information.
+
+This file is then used during the enrichment of the measures data to include those attributes.
+# Running
+## Containerised
+Run the script from the container:
```
marimiyachi: Could you add a note about prereqs here? I think it's just installing Docker.

claims-related/scripts/single_source_conversion_helpers.py
```diff
@@ -0,0 +1,334 @@
...
+DX_CODE_X_CATEGORY = ['DX_CODE_X']
+ENC_PROC_CODE_CATEGORY = ['ENCOUNTER_CODE', 'PROC_CODE']
+QUALITY_CODE_CATEGORY = ['PD_Exe', 'PD_Exl', 'PN_X', 'PN']
+
+# Descriptions of these options can be found at
+# https://cmsgov.github.io/qpp-submissions-docs/schemas# under "Single-Performance Rates".
+PERFORMANCE_OPTIONS = {
+    'PN_X': 'performanceNotMet',
+    'PN': 'performanceMet',
+    'PD_Exe': 'performanceException',
```
marimiyachi: These fields should map to `eligiblePopulationException` and `eligiblePopulationExclusion` instead.

claims-related/scripts/single_source_conversion_helpers.py
```diff
@@ -0,0 +1,334 @@
...
+import numpy as np
+import pandas as pd
+
+DX_CODE_CATEGORY = ['DX_CODE']
+DX_CODE_X_CATEGORY = ['DX_CODE_X']
+ENC_PROC_CODE_CATEGORY = ['ENCOUNTER_CODE', 'PROC_CODE']
+QUALITY_CODE_CATEGORY = ['PD_Exe', 'PD_Exl', 'PN_X', 'PN']
+
+# Descriptions of these options can be found at
+# https://cmsgov.github.io/qpp-submissions-docs/schemas# under "Single-Performance Rates".
```
marimiyachi: This link should be updated: https://cmsgov.github.io/qpp-submissions-docs/measurements#single-performance-rate-measurements

measures/measures-schema.yaml
```diff
@@ -287,18 +287,17 @@ definitions:
...
       - required: ['diagnosisCodes']
-    required: [submissionMethods]
 
   performanceOption:
     type: object
     properties:
       submissionMethods: { $ref: #/definitions/qualityCodesSubmissionMethods }
       optionType:
-        enum: [performanceMet, performanceNotMet, performanceExclusion]
+        enum: [performanceMet, performanceNotMet, performanceExclusion, performanceException]
```
marimiyachi: Same comment here around `eligiblePopulationExclusion/Exception`

measures/measures-schema.yaml
```diff
@@ -287,18 +287,17 @@ definitions:
     anyOf:
       - required: ['procedureCodes']
       - required: ['diagnosisCodes']
-    required: [submissionMethods]
```
marimiyachi: Why is this not required anymore? Or is what you've added only applicable to claims? I'm forgetting the original intention here - but is it possible that registry measures use different codes?

claims-related/scripts/single_source_conversion_helpers.py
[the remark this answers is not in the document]
philipdickinson: Could you elaborate @marimiyachi ? These names are still what exists in the schema. Where are you saying they have been updated?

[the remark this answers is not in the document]
marimiyachi: Yes (and we should update them in the schema too - the schema wasn't actually being used which is why it got out of date) - these should map the fields that the API accepts: https://cmsgov.github.io/qpp-submissions-docs/measurements#single-performance-rate-measurements

measures/measures-schema.yaml
philipdickinson: The allowed submission methods are an attribute of measures, not of particular performance or eligibility options. @akegan do you have any thoughts?

akegan: they are all claims specific, registries do use different codes, but they are not contained in the data at all right now. If we ant to keep submissionMethods as required, we would have to add to performance options too

philipdickinson: @akegan are you saying that different eligibility options can have different permitted submission methods?

claims-related/scripts/single_source_conversion_helpers.py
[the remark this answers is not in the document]
philipdickinson: It sounds like maybe we should hold off on merging this until you get a chance to update the schema. I'm reluctant to make changes without knowing whether they will affect the integrity of the measures data.

measures/measures-schema.yaml
marimiyachi: @philipdickinson you're PR can update these to be `performanceMet, performanceNotMet, eligiblePopulationExclusion, eligiblePopulationException` and then the other code can map correctly to these new names.

claims-related/scripts/data_load.py
```diff
@@ -0,0 +1,88 @@
...
+        'data_element_name': 'str',
+        'gender': 'str',
+        'measure': 'str',
+        'modifier': 'str',
+        'place_of_service': 'str',
+        'reporting_method': 'str'
+    }
+
+    with open(excel_path, 'rb') as fname:
+        single_source = pd.read_excel(fname, sheetname='Codes', dtypes=column_dtypes)
```
tetraptych: Can you replace the `sheetname='Codes'` with `sheetname=-1`? This sheet has a new name in the 2018 Excel file. Thanks!

akegan:
already discussed with philip, but I think we'll need to add submissionMethod to eligibilityOptions too. We can then followup after this PR about a potential better long-term strategy

philipdickinson:
Added @tetraptych 's spreadsheet parsing change, and then added submission methods to all eligibility options (and enforced their requirement in the schema).

PR #110, opened 2018-03-21
QPPA-1733 change value of isInverse for measure 458

ivanang:
previously, isInverse for meaure 458 was false. this PR changes that value to true. it is an inverse measure.

### DIFF
--- a/measures/measures-data.json
+++ b/measures/measures-data.json
@@ -2881,7 +2881,7 @@
     "eMeasureId": null,
     "nqfEMeasureId": null,
     "nqfId": "1789",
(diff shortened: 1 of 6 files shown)


ivanang:
note from @samskeller : 

@kencheeto  ivana's PR  to measures-data (https://github.com/CMSgov/qpp-measures-data/pull/110/files) seems to have changed the integrity hashes of some packages from sha512s to sha1s... not sure why. i checked out her branch and did an `npm install` and _some_ of them changed back, but not all. seems kinda weird. any idea what's going on here/if this is even something to worry about?

PR #114, opened 2018-04-06
Update XML after #108

marimiyachi:
One of the last commits to #108 made some JSON changes that weren't propagated to XML. This PR fixes that.


marimiyachi:
@philipdickinson no worries! Normally you can actually compare the diff but this one was so huge it was impossible to tell!

PR #115, opened 2018-04-10
QPPA-1696 Added General Benchmark and Topped Out Benchmark Updates

sietekk:
This PR updates some general benchmarks and some topped out benchmarks as required in ticket QPPA-1696.


sietekk:
@marimiyachi Yes, at the time only the staging files were updated just for @rockykitamura review. As of now, everything is in the PR and ready for review!

PR #117, opened 2018-04-24
Add additionalProcedureCodes to claims-based measures

tetraptych:
During review of the 2018 measure specifications, it came to our attention that new modified measures, 155 and 226, require multiple sets of procedure codes: one code from list A and one code from list B. This designation was already present in the single source document; this PR adds the designation to the single source parsing script (and regenerates the final `.json` and `.xml` files).

Summary of changes:
- Adds `additionalProcedureCodes` to `eligbilityOption`
- ~Adds 2018 single source file for claims-based measures~
- Populates `additionalProcedureCodes` for claims-based measures using ~2018~2017 single source file
- Regenerates `measure-data.json` and `measures-data.xml`


philipdickinson:
Were there any other changes to the measures JSON? Diff for that file is unfortunately too large to view on the web.

claims-related/scripts/single_source_to_json.py
```diff
@@ -31,10 +31,7 @@
 )
 
 # Default filepaths for all required resources.
-DEFAULT_SINGLE_SOURCE_EXCEL_PATH = (
-    '/home/jovyan/work/data/2017_Claims_IndivMeasures_SingleSourceVersion2_11082017_andrea_edit'
-    '.xlsx'
-)
+DEFAULT_SINGLE_SOURCE_EXCEL_PATH = '/home/jovyan/work/data/2018_Claims_SingleSource.xlsx'
```
akegan: Not sure we want to commit 2018 single source yet, right?

tetraptych:
Update: this PR now includes the same changes as above but applied using the 2017 Single Source Excel file.

measures/measures-schema.yaml
```diff
@@ -300,6 +300,9 @@ definitions:
       procedureCodes:
         type: array
         items: { $ref: #/definitions/codeObject }
+      additionalProcedureCodes:
```
marimiyachi: I see these aren't present in other fields for `eligibilityOption` but would be nice if it's not too hard to add a description here.

PR #118, opened 2018-05-01
Add support for running test suite for claims-based measures

tetraptych:
When we moved the claims-based measure enrichment step into this repository, the unit tests got left behind. 

This PR re-introduces the unit tests for local development, but doesn't add them to CI. Because the tests run inside Docker and (AFAIK) Travis doesn't support container caching, including these tests in CI could result in significantly slower build times. Even if they don't run through Travis, I think it makes sense to have them in the repository for development purposes.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,19 @@
-language: node_js
-node_js: '8'
-cache:
-  directories:
-    - 'node_modules'
+matrix:
+  include:
+    - language: node_js
+      node_js: '8'
+      cache:
+        directories:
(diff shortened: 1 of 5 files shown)


akegan:
@marimiyachi or @kencheeto do you think it would be possible to add these tests to automatically run on travis? Might be good to add in a follow-up PR

tetraptych:
Thanks, @bijujoseph !

tetraptych:
Updated to run the tests in Travis CI.

PR #119, opened 2018-05-15
[QPPA-1933] Prepare measures repo for 2018 performance year measures

kalvinwang:
This PR completes part 1 of 2 of QPPA-1933 so we can begin to add 2018 measures. 

- Move existing 2017 measures to `measures/2017/measures.json` and create new directory for 2018 measures at `measures/2018/measures.json` 
- Update `getMeasuresData` to take an optional year argument (defaulting to 2017)
- Separates the measures-data-spec into 2017 and 2018 test files so it's easier to test differences
- Update the tests to run validation against both using the same existing measures schema file

Coming in Part 2: separate the measures schema for 2017 and 2018

Jira: https://jira.cms.gov/browse/QPPA-1933

Tests: tests updated to loop through all valid performance years, and pass

Reviewer: @marimiyachi


kalvinwang:
@marimiyachi after i merge this, ok if i update QPPA-1606 and 1607 and remove the common language about `If it doesn't exist, make a new directory that will house 2018 measures...`?

test/measures/measures-data-spec.js
```diff
@@ -6,156 +6,159 @@ const path = require('path');
...
 const actualCpcPlusGroups = require('../../util/measures/cpc+-measure-groups.json');
 const actualMeasureSpecificationData = parse(fs.readFileSync(path.join(__dirname, '../../util/measures/measurePDF-Specification.csv'), 'utf8'));
 
-describe('measures data json', function() {
-  const measureIds = _.map(measuresData, 'measureId');
+const performanceYears = [2017, 2018];
+for (const year of performanceYears) {
+  const measuresData = mipsDataFormat.getMeasuresData(year);
+  describe(year + ' measures data json', function() {
+    const measureIds = _.map(measuresData, 'measureId');
```
kalvinwang: This is a poor diff... everything below this line through the end of the file is unchanged other than being indented by two spaces since it's now inside a for..of loop.

test/measures/measures-schema-spec.js
```diff
@@ -7,22 +7,25 @@ const testFilesFolder = 'test/measures/examples/quality/';
 const qualityMeasureExampleFilenames = fs.readdirSync(testFilesFolder);
 
 describe('measures schema validates json', function() {
-  describe('measures-data.json', function() {
-    let capturedStdout = null;
+  const performanceYears = [2017, 2018];
+  for (const year of performanceYears) {
+    describe(year + ' measures-data.json', function() {
```
kalvinwang: This is a poor diff... everything below this line through the end of the file is unchanged other than being indented by two spaces since it's now inside a for..of loop.

scripts/measures/build-measures
```diff
@@ -1,10 +1,11 @@
 #!/usr/bin/env bash
 
+year=2018
```
marimiyachi: Not a huge deal but should this be 2017 until 2018 data is added?

test/measures/measures-data-spec.js
marimiyachi: I wonder if we'll want to create separate spec files for 2017 and 2018. 2018 for example, will have all ACI measures renamed so those tests will need to be rewritten.

marimiyachi: Makes me wonder if we should have different schemas too - there would be a lot of duplication but it would be easier to validate something like oh 2017 has measures that start with ACI and 2018 has measures that start with PI.

kalvinwang: Yeah I think that's a good idea! Will do.

scripts/measures/build-measures
kalvinwang: Yes it should logically... but I just found out I need to add the generated 2018 files in this PR for the tests to run properly, so I'll leave this as 2018-- 2018 data coming tomorrow! :)

test/measures/measures-data-spec.js
kalvinwang: OK I created the separate spec files but separate schemas is a lot more involved so I'll leave that for the next PR

PR #120, opened 2018-05-16
[QPPA-1933] Prepare measures repo for 2018 performance year measures, part 2

kalvinwang:
This PR completes part 2 of 2 of QPPA-1933 so we can begin to add 2018 measures. 

- split schema by peformance years and update docs

Jira: https://jira.cms.gov/browse/QPPA-1933

Tests: tests updated to loop through all valid performance years, and pass

Reviewer: TBD


scripts/measures/build-measures
```diff
@@ -1,11 +1,12 @@
 #!/usr/bin/env bash
 
-year=2018
+defaultPerformanceYear=2018
+performanceYear=${1:-$defaultPerformanceYear} # check if year provided on command line ($1)
```
marimiyachi: Since we won't every be rebuilding measures data for 2017, should this not be an argument and actually just default to 2018? Then we can remove the comment in Constants about editing this in two places.

measures/2018/measures-schema.yaml
```diff
@@ -0,0 +1,351 @@
...
+      category:
+        description: QPP scoring category to which the measure belongs: Improvement Activities, Quality, Advancing Care Information, and Cost.
+        enum: [ia, quality, aci, cost]
+      metricType:
+        description: Type of measurement that the measure requires in order to attest.
+        enum: [boolean, proportion, singlePerformanceRate, multiPerformanceRate, registrySinglePerformanceRate, registryMultiPerformanceRate, nonProportion, cahps, costScore]
+      firstPerformanceYear:
+        description: Year in which the measure was introduced.
+        type: integer
+        default: 2017
```
marimiyachi: Should this still be 2017? Or maybe 2018 instead?

scripts/measures/build-measures
kalvinwang: great point! ok

kalvinwang: hm... actually, I should keep the comment in Constants because its main use case is, when next year comes around, there's a second place where you should change it. I'll rename it from `default` to `current` though. 

removed it as an argument to build-measures!

README.md
```diff
@@ -32,15 +32,17 @@ navigate to `benchmarks/`. Benchmarks data is organized by performance year.
...
 The module can be used with the following pattern:
 ```javascript
 const qppMeasuresData = require('qpp-measures-data');
-const measuresData = qppMeasuresData.getMeasuresData(2018);
-const measuresSchema = qppMeasuresData.getMeasuresSchema(2018);
+const measuresData = qppMeasuresData.getMeasuresData($YEAR);
+const measuresSchema = qppMeasuresData.getMeasuresSchema($YEAR);
 const benchmarksData = qppMeasuresData.getBenchmarksData();
-const benchmark2017Data = benchmarksData[2018];
+const benchmark2017Data = benchmarksData[$YEAR];
```
harrisj: Maybe this variable should be renamed?

kalvinwang: @harrisj haha good catch. i'll fix it in my upcoming pr

PR #125, opened 2018-05-30
[QPPA-1607] Add tests and update ajv, ajv-keywords, eslint, mocha, lodash 

kalvinwang:
- Add ajv-keywords so we can make use of `uniqueItemProperties: ['measureId']`
  - add the duplicate measure validation tests @samskeller suggested in #124 
- Update ajv 
  - migrate 2018 measures-schema from the draft-04 JSON schema spec to the draft-07 spec ([breaking changes](http://json-schema.org/draft-06/json-schema-release-notes.html#q-what-are-the-changes-between-draft-04-and-draft-06) were in 06)
- Update lodash (fixes _.isEqual bug: https://github.com/lodash/lodash/issues/1758)
- Update mocha (`npm audit` showed a critical vulnerability in dependencies)
- Update eslint (`npm audit` showed a low severity vulnerability in dependencies)

Tests: added new tests, passes existing

Jira ticket: https://jira.cms.gov/browse/QPPA-1607 (adding the duplicate measure validation tests associated with 1607 led to the rest of these items)

Reviewer: @samskeller


measures/2018/measures-schema.yaml
```diff
@@ -1,7 +1,8 @@
-id: https://github.com/CMSgov/qpp-measures-data/versions/0.0.1/measures-schema.yaml
-$schema: http://json-schema.org/schema#
+$id: https://github.com/CMSgov/qpp-measures-data/versions/0.0.1/measures-schema.yaml
```
kalvinwang: Do we want to update this url? This url doesn't exist, and I don't know if it ever did. I know we've updated this schema many times and I'm now updating it again soo... @marimiyachi @kencheeto

samskeller: Hmm I don't have much context, but could we just delete it...?

measures/2018/measures-schema.yaml
```diff
@@ -1,7 +1,8 @@
...
+$schema: http://json-schema.org/draft-07/schema
```
samskeller: Curious what's this for? Cause I can't imagine it's making a network call to download the schema when verifying this file...?

kalvinwang: nope no effect, just for a reader

https://spacetelescope.github.io/understanding-json-schema/basics.html#declaring-a-json-schema
```
The $schema keyword is used to declare that something is JSON Schema. It’s generally good practice to include it, though it is not required.

You can also use this keyword to declare which version of the JSON Schema specification that the schema is written to.
```

measures/2018/measures-schema.yaml
kalvinwang: Hm ok. I'll just update it with `https://github.com/CMSgov/qpp-measures-data/blob/master/measures/2018/measures-schema.yaml` since it's apparently a unique identifier and a base url for $ref (though all of our ref urls are relative...) https://spacetelescope.github.io/understanding-json-schema/structuring.html#the-id-property

PR #132, opened 2018-06-15
Fix order of measure sets

marimiyachi:
We were expecting measure sets in alphabetical order. Short term fix is to change the script to expect the non-alphabetical order of the csv. Long term fix is to not hardcode indexes but just look for column names directly.


kencheeto:
failing test that needs to be updated i think, but lgtm otherwise!

PR #136, opened 2018-07-02
Feature/qppsf 2594 is icd impacted is topped out program fields

NateTheGreatt:
Added two new fields to the measures data:

- isIcdImpacted
- isToppedOutByProgram

Note: Tests not passing, awaiting this PR for test data cleanup:
https://github.com/CMSgov/qpp-measures-data/pull/135


KyleApfel:
@NateTheGreatt I'm noticing a lot of keys changed under "description" and "title" that shouldn't have updated text and could be a result of Excel mutating the string. Please go back and make sure the .json and .xml files generated are only modifying the `isIcdImpacted` and `isToppedOutByProgram` fields.

You may need to revert to the old csv file to restore the changes.

NateTheGreatt:
@KyleApfel confirmed, I see those alterations. Thank you, will address.

PR #137, opened 2018-07-05
[QPPA-2054] Update 2018 quality measures, part 2

kalvinwang:
This PR updates the quality measures strata source CSV and the quality measures JSON output. It's a followup to #135.

- Fixes mystery � characters in quality measure multiperformancerate strata (shoutout to @sarahnw):

Jira: https://jira.cms.gov/browse/QPPA-2054

Tests: tests pass

Reviewer: @kencheeto


staging/2018/measures-data-quality.json
```diff
@@ -7652,10 +7652,6 @@
       {
         "name": "2-3",
         "description": "Children in the eligible population who turned 3 during the performance period ( > 2 years to 3 years of age) who were screened for risk of developmental, behavioral and social delays using a standardized tool with interpretation and report within 12 months preceding or on their birthday"
-      },
-      {
-        "name": "overall",
-        "description": "All children who turned 1, 2, or 3 during the performance period who were screened for risk of developmental, behavioral and social delays using a standardized tool with interpretation and report within 12 months preceding or on their birthday"
```
kencheeto: this is supposed to be removed?

kalvinwang: good catch! @sarahnw the last line of the previous strata CSV has been removed in this new source CSV, and it's the overall strata for measure 467. was that intentional or should it be kept?

sarahnw: that should be removed! received clarification on that measure from PIMMS through a service now ticket. overall is basically just the performance rate which the API calculates so no need to report it separately.

PR #144, opened 2018-07-25
Add supplemental fields for claims-based quality measures to measures-data.json

tetraptych:
This PR follows-up [this PR](https://github.com/CMSgov/qpp-measures-data/pull/143) by actually adding the new fields for claims-based quality measures to the final `measures-data.json` and `.xml` files.

This PR is not ready for merging for the following two reasons. If one of the reviewers has the bandwidth, I'd deeply appreciate help resolving them! 

1. The tests in `import-quality-measures-spec.js` are failing with the following message:
```
AssertionError: Unexpected properties on measureId 001: expected [ Array(22) ] to have the same members as [ Array(24) ]
```
Seems obvious enough -- since (two) new fields have been added, the arrays no longer match. Is there some way this test can be more flexible for quality measures with the claims reporting option?

2. I have not bumped the version numbers or cut a release. I'm hopeful that someone more familiar with this process can push the necessary changes to this branch before merging.

Cheers!


tetraptych:
Thanks for your thorough review, @kalvinwang ! I'll be going through your comments and addressing them on this branch. I'll open a new PR when I think things are ready for a re-review.

tetraptych:
Re: 1. above: I updated the 2018 schema to not require `submissionMethods` in `eligibilityOptions` and `performanceOptions` and also stopped populating these fields in `measures-data.json`.

tetraptych:
Re 2: above (measures with hundreds of "code" objects): there are indeed examples with keys besides `"code"`, the most common being `"modifiers"` and `"placesOfService"` [to indicate that the CPT/HCPCS code only applies if a particular modifier code is also present or if the procedure was performed in a certain setting (inpatient / outpatient / emergency room / hospice / etc.) ].

Sadly these fields can have different values across codes in the same measure -- one common scenario for some measures is that the code indicating `performanceNotMet` is the same as the code for `performanceMet`, but with a different modifier.

kalvinwang:
@tetraptych thanks for the detailed explanation! 

separately, just curious, why close this PR and open a new one? isn't it easier/cleaner to just keep this same one open so all the context is on the same PR?

tetraptych:
Definitely wasn't my intention to have two PRs for this! As a side-effect of the rebase onto master, the local and remote had different commits. Apparently the repo was unhappy about my use of force-push to update the remote, since it closed the PR and disabled the "Reopen and comment" button below.

kalvinwang:
Weird! OK gotcha.

PR #145, opened 2018-07-30
eCQM UUID updates

saquino0827:
- Adds back in enrichment for 2018 for the purpose of updating quality measure UUIDs
  - I removed updating the measure specification as that was set to 2018 already.
  - I updated ACI -> PI in the enrichment step and used the PI relations data.
  - Inverse measures, CPC+ measure groups and additional stratifications still use 2017. I do not know what 2018 updates they require or if they can be removed.


kencheeto:
@saquino0827 looks like there's a couple test failures - let me know if this pr is ready for review. thanks!

scripts/measures/2018/get-strata-uuids-from-ecqm-zip-2018.js
```diff
@@ -0,0 +1,171 @@
...
+      if (emeasureid === '145') {
+        console.warn('WARNING: CMS145v6 has one numerator but two initial populations and needs to be added manually - see /tmp/ecqm/EC_CMS145v6_NQFXXXX_CAD_BB.zip');
+        return;
+      }
+      if (emeasureid === '160') {
+        console.warn('WARNING: CMS160v6 has one numerator but two initial populations and needs to be added manually - see /tmp/ecqm/EC_CMS160v6_NQF0712_Dep_PHQ9.zip');
+        return;
+      }
+      if (emeasureid === '347') {
+        console.warn('WARNING: CMS347v1 has one numerator but three initial populations and needs to be added manually - see /tmp/ecqm/EP_EC_CMS347v1_NQFXXXX_Statin_Therapy_Prev.zip');
```
kencheeto: @saquino0827 do these edge cases still exist? i wrote the first version of this script for 2017 measures, wondering if the 2018 ecqm zips are the same

scripts/measures/2018/enrich-measures-data.js
```diff
@@ -0,0 +1,164 @@
...
+  mergeGeneratedEcqmData(measures);
+  addRequiredRegistrySubmissionMethod(measures);
+  enrichClaimsRelatedMeasures(measures);
+  return JSON.stringify(measures, null, 2);
+};
+
+/**
+ * Will add extra metadata to PI measure that are not directly available
+ * in machine inferable format at https://qpp.cms.gov/api/v1/aci_measures
+ * After this function executes, an ACI measure will have reporting category and substitutes.
```
kencheeto: could you update ACI here and in other places in this pr to read PI?

saquino0827: Will do, I thought I got all the ACI references but I missed a some. Good catch.

scripts/measures/2018/get-strata-uuids-from-ecqm-zip-2018.js
saquino0827: If I understand it, they do exist. I thought that if they had multiple strata(numerator) values, then they could not be extracted from the initial script and needed to be handled separately. Whenever I ran the script for the current zip file, the 4 eMeasureId's would always fail for the same reason.

saquino0827:
@kencheeto I don't believe the format changed much at all. The most I saw were UUID updates, but the initial get-strata-uuids-from-ecqm-zip script pretty much worked after some minor tweaks. I'll go back and double checked any comments I left out.

saquino0827:
I forgot to mention that CMS347v1 is still undergoing a change. I had to ask if my data was correct because it has multiple strata but is listed as a singlePerformanceRate. I changed it to multiPerformanceRate but then asked if that is correct. This is what caused the test failures.

kencheeto:
@saquino0827 is it correct now then? looks like there are some conflicts now w/master, lgtm after rebasing

saquino0827:
@kencheeto I've fixed the conflicts. I've left some 2017 data in that can be removed if needed, such as inverse-measures and additional stratifications. I don't have updated data for these and don't know if they are required.

kencheeto:
@saquino0827 ok, no blockers to merge from my perspective then! go ahead

sarahnw:
@saquino0827 - in this PR you changed measure 226 to weightedAverage instead of overall, can you provide more info on that? The measures team says it should be overall. Let me know!

saquino0827:
@sarahnw I checked on this but it was just my mistake. I thought I had corrected my data. I'll resubmit a PR with a fix shortly.

PR #146, opened 2018-07-31
upgrade adm-zip to fix vulnerability

kencheeto:
screen shot 2018-07-31 at 10 43 41 am

https://nvd.nist.gov/vuln/detail/CVE-2018-1002204

Not a likely attack vector for us since we manually process zip files, but upgrades adm-zip anyway to resolve the sec warning that GitHub displays.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "qpp-measures-data",
-  "version": "1.6.1",
+  "version": "1.6.2",
   "description": "Quality Payment Program Measures Data Repository",
   "repository": {
     "type": "git",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "qpp-measures-data",
-  "version": "1.6.1",
+  "version": "1.6.2",
```
kalvinwang: should we not bump the version number? the readme says to always bump when measures data is changed, and otherwise optional; this change seems so invisible that we should skip a version bump (unless it's valuable to have a release for every change of any kind?)

kencheeto: i should rtfm! good catch

PR #148, opened 2018-08-07
increase test timeout for link checking

kencheeto:
the test we merged recently to check that the measure specification links are valid failed because of a timeout, so I'm increasing the timeout here by 50%.

this test only runs on CI during cron runs, once a day.

failing run here:
https://travis-ci.org/CMSgov/qpp-measures-data/jobs/412016666

### DIFF
--- a/test/measures/measures-specification-spec.js
+++ b/test/measures/measures-specification-spec.js
@@ -28,8 +28,8 @@ function checkUrl(s) {
 // this will run once a day on travis
 if (process.env.TRAVIS_EVENT_TYPE === 'cron') {
   it('has valid specification links', function() {
-    this.timeout(12000); // 8 seconds timeout.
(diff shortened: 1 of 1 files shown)


test/measures/measures-specification-spec.js
```diff
@@ -28,7 +28,7 @@ function checkUrl(s) {
 // this will run once a day on travis
 if (process.env.TRAVIS_EVENT_TYPE === 'cron') {
   it('has valid specification links', function() {
-    this.timeout(8000); // 8 seconds timeout.
+    this.timeout(12000); // 8 seconds timeout.
```
kalvinwang: i would just delete the comment so it can't be wrong, it doesn't seem useful :p

kencheeto: duh. fixed!

PR #149, opened 2018-08-07
Update enrichment to use 2018 claims-based quality measures data

tetraptych:
Remove 'submissionMethods' from eligibilityOptions and performanceOptions

Rebase on modified source branch

Add supplemental fields for claims-based quality measures to measures-data.json


tetraptych:
This PR incorporates the changes made in #144 in light of the re-addition of the enrichment script for quality measures introduced in #145. It also re-generates the measures JSON file using the 2018 definitions for claims-based quality measures. 

Still reviewing what new test(s) might be useful in `merge-measures-data-spec.js` or `validate-data-spec.js`.  

(CC: @kalvinwang)

kalvinwang:
@tetraptych thanks for addressing most of the comments from the previous PR (#144)! The remaining one was about tests for import-quality-measures. I'm confused actually, what happened to the changes in import-quality-measures.js, etc from #144? Are those intentionally left out of this PR? #144 was closed not merged so it seems like those changes are still needed?...

tetraptych:
The original PR was intended to add the enrichment step `enrichClaimsRelatedMeasures` from `enrich-measures-data.js` in 2017 to the 2018 workflow. This file didn't exist in `2018/`, so the original PR added the enrichment step to `import-quality-measures.js`. 

PR #145 made this much easier by creating an explicit enrichment script in `2018/` with all the previous 2017 enrichment steps -- including `enrichClaimsRelatedMeasures`. This enrichment step was still using the 2017 file, so the main update here is to point it at the 2018 file instead.

Looking over the changes, maybe it's clearer sense to keep the `qpp-single-source-2018.json` filename and update the enrichment script to use that path instead of `qpp-single-source`.

PR #151, opened 2018-08-16
[QPPSF-2735] Add 2018 QCDR Measures

NateTheGreatt:
This PR takes the 2017 QCDR measures import script and adds it to the /2018 scripts folder, and adds it into the build-measures bash script. It is invoked after the previous measures have been merged together because there is custom conflict-resolution merging logic in the QCDR measures import script which has the potential to modify measures that were previously imported. However, currently that logic is not hit. Should it be removed for clarity? If so, I can modify and fit the QCDR import script into the current merge script.


kalvinwang:
@NateTheGreatt what's the specific conflict-resolution logic you're referring to that isn't hit, and perhaps could be removed? also do you have a sense of why it isn't hit / what's changed about measures-data or the new 2018 qcdrs such that it's no longer hit?

kalvinwang:
(chatted with @NateTheGreatt on Slack, he was referring to `scripts/measures/2018/import-qcdr-measures.js:206` and my answer was "yes we should keep it because once the qcdr measures are imported into measures-data (which your pr is doing) then the qcdr measureIds will be present in measures-data, and re-importing changes to the qcdr csv (last year there were like… a dozen, all named “final”) will require this function")

scripts/measures/2018/import-qcdr-measures.js
```diff
@@ -79,10 +79,17 @@ const config = {
     primarySteward: 20
     // `metricType` is a sourced field but not represented here since it maps from
     // multiple columns-- you can find it by searching in the code below
+  },
+  special_fields: {
```
kalvinwang: what do you think of calling this `referenced_fields`, if that accurately captures it, and changing sourced_fields to `mapped_fields`? also, can you add a new comment below `'source_fields' are fields which...` at the top of this file, describing `referenced_fields` (or whatever it's named)?

scripts/measures/2018/import-qcdr-measures.js
```diff
@@ -120,7 +129,7 @@ const addMultiPerformanceRateDetails = function(newMeasure, record, qcdrStrataNa
   newMeasure['strata'] = [];
   _.each(strata, function(stratum, index) {
     if (_.isUndefined(qcdrStrataNames[measureId])) {
-      throw TypeError('Missing strata for ' + measureId + '. Should' +
+      throw TypeError('Missing strata for ' + measureId + '. Should ' +
```
kalvinwang: nice catch

NateTheGreatt:
@kalvinwang I have seen that happen in a previous PR before. It can be circumvented by recreating the PR, if that's preferable.

kalvinwang:
No need, good to know for next time!

util/measures/2018/qcdr-measures-strata-names.json
```diff
@@ -0,0 +1,42 @@
+{
+  "AHSQC6":["overall","<=10cm",">10cm"],
```
rockykitamura: @NateTheGreatt @kalvinwang Are the strata names suppose to match the 2017 QCDR measure strata names when it's the same measure in both years?  I'm finding a few mismatches for example in 2017 for measure AHSQC6 (https://github.com/CMSgov/qpp-measures-data/blob/master/util/measures/2017/qcdr-measures-strata-names.json) the strata names were: `"AHSQC6": ["hernia", "overall", "hernia>10cm"]`

Should we use similar notation `"hernia>10cm"` and `"hernia<=10cm"`

NateTheGreatt: I tried to align the names with the strata names that were embedded inside of the `quality-strata.csv` file in this ticket: https://jira.cms.gov/browse/QPPA-2098

rockykitamura: Is there any benefit to keeping the strata names consistent (when possible) from year to year?  For example, do we know if data from the same strata name will be compared across years programatically?

sarahnw: yes - if the strata itself doesn't change then we should keep them the same. we manually name these and I thought I had gone through and aligned them. are there other ones that aren't matched? let me know and I can re-review the document.

sarahnw: In the case of AHSQC6, for 2018 we were told rate 1 was the overall, which is 2017 was named "hernia" so in this particular case the names changed for 2 of the 3 rates.

NateTheGreatt: @sarahnw There is also a hard-limit of 20 characters for the strata names set in the code, currently. The strata names for measureId `PP4` in the previously mentioned `quality-strata.csv` file breach this validation rule. @kalvinwang should this be increased?

sarahnw: oh interesting I didn't know that. we can change the names! that measure was trickier to name... what about:
counselalcohol
counseltobacco
counselsubstance
counselactivity

kalvinwang: @NateTheGreatt hm not sure why that 20char limit exists. i think we can leave that for a future pr since the current names are fixed now. thanks for the heads up

kalvinwang:
@NateTheGreatt oh also, same thing as on your other PR, #152-- you'll want to update the version number in package.json in this PR since you're updating measures-data.json (see https://github.com/CMSgov/qpp-measures-data/blob/master/CONTRIBUTING.md#versioning-publishing-and-creating-new-releases) 

If you're merging/releasing this one before #152, you'll probably want to set this PR's version to 1.7.1 and the other PR to 1.7.2 (or 1.8.0 and 1.8.1, since adding QCDR measures seems worth a larger bump)

Or, if you're able to merge both of these PRs right after one another, you could skip updating package.json in this PR and add the release note for this PR into the one for #152.

PR #163, opened 2018-10-22
Fix strata names

saquino0827:
- Update to quality-strata.csv in util/measures/2018
- Updates enrich-measures-data.js to read the new quality-strata.csv and update strata names accordingly.
    - Some measureId's were not updated in quality-strata.csv because it was determined that they are not eCQMs as they do not have eMeasureId's (Verified with @sarahnw ).
- Updated measure-data.json file with strata names.
- Update package.json version number. I cannot publish this however.


NateTheGreatt:
The pre-existing strata code @`import-quality-measures.js:addMultiPerformanceRateStrata()` also needs to be removed.

saquino0827:
@NateTheGreatt I have fixed both '009' and '239'. I have also added a new test for strata names under 'measure-data-spec.js'

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "qpp-measures-data",
-  "version": "1.8.9",
+  "version": "1.9.0",
```
NateTheGreatt: Because this is a bugfix let's just bump the patch version: `1.8.10`

saquino0827: Got it. Will change.

scripts/measures/2018/enrich-measures-data.js
```diff
@@ -128,6 +130,27 @@ function mergeGeneratedEcqmData(measures) {
...
+  qualityStrataCsv.forEach(function(strata, csvIndex) {
+    if (strata[6]) {
+      if (strata[0].length === 1) strata[0] = '00' + strata[0];
+      if (strata[0].length === 2) strata[0] = '0' + strata[0];
+      measures.forEach(function(qppItem, qppIndex) {
+        if (qppItem.category !== 'quality') return;
+        if (qppItem.eMeasureId == null) return;
+        if (qppItem.measureId === strata[0]) {
+          measures[qppIndex].strata.forEach(function(measureStrata, strataIndex) {
+            if (measureStrata.eMeasureUuids && measureStrata.eMeasureUuids.numeratorUuid === strata[6]) {
```
kalvinwang: good call on checking if eMeasureUuids exists. No need to change this if you never use lodash, but FYI you can also replace the left side of this with `_.get(measureStrata, 'eMeasureUuids.numeratorUuid')`

scripts/measures/2018/enrich-measures-data.js
```diff
@@ -128,6 +130,27 @@ function mergeGeneratedEcqmData(measures) {
...
 }
 
+function addQualityStrataNames(measures) {
+  const qualityStrataCsv = parse(fs.readFileSync(path.join(__dirname, '../../../util/measures/2018/quality-strata.csv'), 'utf8'));
+  qualityStrataCsv.forEach(function(strata, csvIndex) {
+    if (strata[6]) {
+      if (strata[0].length === 1) strata[0] = '00' + strata[0];
+      if (strata[0].length === 2) strata[0] = '0' + strata[0];
+      measures.forEach(function(qppItem, qppIndex) {
+        if (qppItem.category !== 'quality') return;
```
kalvinwang: suggestion: you could combine these three lines into

```
if (qppItem.category !== 'quality' || 
_.isNull(qppItem.eMeasureId) || 
qppItem.measureId !== strata[0]) 
  return;
```

package.json
kalvinwang: good call!

scripts/measures/2018/enrich-measures-data.js
```diff
@@ -128,6 +130,27 @@ function mergeGeneratedEcqmData(measures) {
   });
 }
 
+function addQualityStrataNames(measures) {
```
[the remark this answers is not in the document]
saquino0827: Will do.

scripts/measures/2018/enrich-measures-data.js
```diff
@@ -128,6 +130,27 @@ function mergeGeneratedEcqmData(measures) {
...
+  const qualityStrataCsv = parse(fs.readFileSync(path.join(__dirname, '../../../util/measures/2018/quality-strata.csv'), 'utf8'));
+  qualityStrataCsv.forEach(function(strata, csvIndex) {
+    if (strata[6]) {
```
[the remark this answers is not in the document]
saquino0827: 👍

scripts/measures/2018/enrich-measures-data.js
saquino0827: Okay. I'll look into this

saquino0827:
@kalvinwang I went with your suggestions and made the changes. I will push these changes up if everything looks good.

scripts/measures/2018/enrich-measures-data.js
```diff
@@ -128,6 +130,29 @@ function mergeGeneratedEcqmData(measures) {
   });
 }
 
+/*
+ * A function that uses an identifier in `quality-strata.csv` to map a numerator uuid to a strata.
```
kalvinwang: Ah I see. The identifier is the uuid right? If so, it seems confusing to reference it separately. How about combining these two lines into:

```Uses numeratorUuid field as a common id to map each strata name (only in `enriched-measures-data-quality.json`) to a particular strata (in `quality-strata.csv`)```

saquino0827: 👍

sarahnw:
@saquino0827 measure 226 now has 2 strata named overall. only the second should have that name.

sarahnw:
the third strata is now called "tobacco" as noted in the spreadsheet on the ticket.

kalvinwang:
@saquino0827 while looking into @sarahnw's comment above, I realized that the output measures-data.json seems to be really off. E.g. it look like many of the strata for different measures all ended up being output as strata inside measure 009, including the overall/intervention/tobacco strata Sarah's referencing above. Could you please take another look and figure out what's happening? Thanks!

sarahnw:
@saquino0827 can you look at this to fix as soon as possible? It looks like the strata names are outside of the strata and therefore still not usable. It's breaking the dev preview since developers can't submit any of these measures until the names are back in correctly.

PR #168, opened 2018-10-26
QPPSF-2996 Remove EMA 2018 clinical clusters

rockykitamura:
We are removing the EMA clinical clusters for 2018 (https://jira.cms.gov/browse/QPPSF-2996) but keeping the specialty sets.


rockykitamura:
So, we don't want to remove the 208 EMA clinical clusters after all.  Sorry for the confusion.

PR #171, opened 2018-10-31
Fix Strata name for measure 226

saquino0827:
- Removes the duplication of the name `overall` with the name `tobacco`.
@sarahnw This is to address the comment you made in my strata name PR that was already merged.


sarahnw:
First strata: "screenedForUse" description "Percentage of  patients aged 18 years and older who were screened for tobacco use one or  more times within 24 months"

Second strata: "overall" description: "Percentage of patients aged 18 years and   older who were screened for tobacco use and identified as a tobacco user who   received tobacco cessation intervention"

Third strata: "tobacco" description: "Percentage of patients aged 18 years and   older who were screened for tobacco use one or more times within 24 months   AND who received tobacco cessation intervention if identified as a tobacco user"

@saquino0827 - can you double check the descriptions with the UUIDs you are assigning? It looks like you've just moved what was the old second rate to be the third in some cases, but that's not correct.

saquino0827:
Thank you for pointing that out @sarahnw . I fixed the description issue.

kalvinwang:
@saquino0827 just checking, did you already check the other descriptions against the UUIDS per sarah? if not could you, just in case? it's unfortunate it's such a manual process :(

saquino0827:
@kalvinwang yes I checked with Sarah and she said it was correct.

PR #172, opened 2018-11-02
QPPA 2447 : Run script to ingest updated 2018 MIPS benchmarks

shebryant:
HCQAR made an update to the 2018 benchmarks file and the updated file needed to be ingested in order to update benchmarks.

Test plan:
Unit tested

Associated JIRA tickets:
https://jira.cms.gov/browse/QPPA-2447

Reviewers: @lamroger, @samskeller


shebryant:
@samskeller Chatted with @sarahnw. From what I gather, there will be a separate ticket, but I believe Sarah will be adding a comment for more clarification,

shebryant:
HCQAR made an update to the 2018 benchmarks file and the updated file needed to be ingested in order to update benchmarks.

sarahnw:
@samskeller good catch, I've compiled the latest quality changes in a new ticket QPPA-2562

PR #175, opened 2018-11-06
Reporting Stratum CMS

saquino0827:
- Adds Reporting Stratum CMS UUIDs into the measure's that require them.


scripts/measures/2018/enrich-measures-data.js
```diff
@@ -130,6 +109,28 @@ function mergeGeneratedEcqmData(measures) {
   });
 }
 
+/**
+ * Adds in each SubPopulation's stratification UUIDs
+ * This JSON document used to derive this is generated using get-stratifications.js
+ * Still uses 2017 additional stratifications from 2017. This is subject to change.
+ */
+function enrichStratifications(measures) {
```
kalvinwang: did you make any changes here other than moving it down? no? just checking if i should be reviewing it more closely

saquino0827: I only moved the code down in the steps. I moved them down because not all the measures are completely created with strata initially. And stratifications are like a `sub-strata` as they belong inside some specific measure's strata.

PR #177, opened 2018-11-07
QPPA 2563: Add Exclusion Field to Specified Measures

shebryant:
This PR adds an "exclusion" field to some specified PI measures to store the id of exclusion measures, measures that can be submitted instead of the coordinating base measure.

Unit tested.

Associated JIRA tickets: https://jira.cms.gov/browse/QPPA-2563

Reviewers: @kalvinwang , @samskeller


test/measures/2018/measures-data-spec.js
```diff
@@ -33,7 +33,11 @@ describe(year + ' measures data json', function() {
       measuresData
         .filter(m => m.category === 'pi')
         .forEach(m => {
-          generated[m.measureId] = {reportingCategory: m.reportingCategory, substitutes: m.substitutes};
+          if (m.exclusion) {
```
kalvinwang: super minor, you could duplicate less code by keeping the original line, then
```
if (m.exclusion) {
  generated[m.measureId].exclusion = m.exclusion'
}
```

PR #229, opened 2018-12-11
Update claims-related files in preparation for 2019

tetraptych:
Add 2019 Single Source csv for claims-based quality measures


claims-related/scripts/data_load.py
```diff
@@ -66,8 +66,18 @@ def excel_to_csv(excel_path):
         'reporting_method': 'str'
     }
 
-    with open(excel_path, 'rb') as fname:
-        single_source = pd.read_excel(fname, sheetname=-1, dtypes=column_dtypes)
+    # Try to use the sheet named "Codes," otherwise default to the last sheet in the file.
```
marimiyachi: Should this just error if the proper sheet name doesn't exist? Why use the last one?

tetraptych: Agreed!

PR #230, opened 2018-12-12
[QPPA-2658] Remove measure 226 from 2018 benchmarks

kalvinwang:
Remove measure 226 from 2018 benchmarks

I manually removed the measure from the CSV and JSON layer files (the more correct way would be to run parse-benchmark-data.js on the CSVs, but then again the really correct way would be to run parse-benchmark-data automatically as part of build-benchmarks)

Then ran `npm run build:benchmarks` per CONTRIBUTING.md to generate the resulting 2018.json benchmarks file

Jira ticket: https://jira.cms.gov/browse/QPPA-2658

Reviewer: @lamroger


sarahnw:
@kalvinwang I think you mean measure 226 not 236!

sheldon-b:
First time reviewing a qpp-measures-data PR so this is all new to me. Were there a series of steps you did to produce these changes? It looks like the changes are probably automated, i.e. processing the csv file that Sarah attached to the Jira ticket

kalvinwang:
@sarahnw good point, I've updated the PR title/description :P 

@sheldon-b this is actually something that I don't think we've really done before (as the QPPA team), remove a benchmark manually, so not really a common category of measures-data PRs to the extent there are categories. I'll flesh out the PR description to explain what I did, good idea

benchmarks/2018.json
```diff
@@ -5071,35 +5017,19 @@
     "benchmarkYear": 2016,
     "performanceYear": 2018,
     "submissionMethod": "cmsWebInterface",
+    "isToppedOut": false,
```
lamroger: do you want these changes

benchmarks/2018.json
```diff
@@ -5071,35 +5017,19 @@
...
     "deciles": [
       0,
       30,
       30,
       40,
       50,
-      60,
+      66,
```
lamroger: here too

staging/2018/benchmarks/csv/000_benchmarks_base.csv
```diff
@@ -160,9 +160,6 @@ Functional Status Change for Patients with General Orthopaedic Impairments,223,R
 Melanoma: Overutilization of Imaging Studies in Melanoma,224,Registry/QCDR,Process,Y,--,--,--,--,--,--,--,100,Yes,
 Radiology: Reminder System for Screening Mammograms,225,Claims,Structure,Y,94.20 - 98.22,98.23 - 99.67,99.68 - 99.99,--,--,--,--,100,No,
 Radiology: Reminder System for Screening Mammograms,225,Registry/QCDR,Structure,Y,--,--,--,--,--,--,--,100,Yes,
-Preventive Care and Screening: Tobacco Use: Screening and Cessation Intervention,226,Claims,Process,Y,96.65 - 98.59,98.60 - 99.61,99.62 - 99.99,--,--,--,--,100,Yes,
-Preventive Care and Screening: Tobacco Use: Screening and Cessation Intervention,226,EHR,Process,Y,72.72 - 82.40,82.41 - 87.55,87.56 - 90.50,90.51 - 92.74,92.75 - 94.63,94.64 - 96.52,96.53 - 98.51,>= 98.52,No,
-Preventive Care and Screening: Tobacco Use: Screening and Cessation Intervention,226,Registry/QCDR,Process,Y,83.72 - 88.38,88.39 - 91.10,91.11 - 93.34,93.35 - 95.44,95.45 - 97.35,97.36 - 99.00,99.01 - 99.99,100,No,
```
lamroger: there were 4 deleted above but only three here. does there need to be another one deleted

lamroger: missing is cmsWebInterface

benchmarks/2018.json
kalvinwang: Ah... nope, it looks like this was changed in https://github.com/CMSgov/qpp-measures-data/pull/227/files in the json, but not the csv, so when I reran build-benchmarks these changes were overwritten. I should've reviewed the original PR more closely!

benchmarks/2018.json
kalvinwang: Same as above

staging/2018/benchmarks/csv/000_benchmarks_base.csv
kalvinwang: nope, it looks like this the problem is cmsWebInterface shouldn't have been removed above-- that was added in https://github.com/CMSgov/qpp-measures-data/pull/227/files in the json, but not the csv, so when I reran build-benchmarks these changes were overwritten

kalvinwang:
@lamroger thanks for catching these! i'm making a quick fix for now and shad is creating a separate pr to fix the source

PR #233, opened 2018-12-14
QPPSF-3093 DM Composite WI Benchmark Fix

shadford:
Fixed merge issue with DM Composite WI Benchmarks

---

### Summary

Fixed merge issue with DM Composite WI Benchmarks, file was replaced by https://github.com/CMSgov/qpp-measures-data/pull/230

### Test plan:
Things that have to pass. These are things that the reviewer(s) should be able to reproduce. This can also include a list like:

* [ ] Migrations are backwards-compatible
* [X] Unit tested
* [X] Verified working locally
* [ ] Documented

### Associated JIRA Tickets

* https://jira.cms.gov/browse/QPPSF-3093

### Reviewers

@kalvinwang @lamroger @NateTheGreatt @bijujoseph @KyleApfel @rockykitamura @marimiyachi @samskeller @kyeah @sheldon-b @shebryant


staging/2018/benchmarks/json/001_benchmarks_wi.json
```diff
@@ -153,21 +162,21 @@
...
+    "performanceYear": 2018,
+    "submissionMethod": "cmsWebInterface",
+    "isToppedOut": false,
+    "deciles": [
+      0,
+      30.00,
+      30.00,
+      40.00,
+      50.00,
+      60.00,
```
kalvinwang: This line is the only merge issue from #230

shadford:
Hey Kalvin,
After talking with Rocky we decided to add back in the 'isToppedOut: false' back in as before to keep everything consistent. 

I add the csv, however I'm running into issues running the parse-benchmark-data.js if you have time tomorrow could you help me out with this?

staging/2018/benchmarks/json/001_benchmarks_wi.json
shadford: Talked with Rocky, the is isToppedOut: false should not have been removed. Added it back in to keep things consistent.

kalvinwang:
@shadford oh, ok that makes sense then. In that case you'll also want to add them back in the intermediate files where they were removed in #227 (in staging/2018/benchmarks/json/001_benchmarks_wi.json)

Parse-benchmark-data.js doesn't seem to be in working condition and I've created a ticket here to track fixing, so no worries about making use of it! https://jira.cms.gov/browse/QPPA-2693 you can just use the build script to generate the final output (2018.js) from the intermediate files, and manually edit the csv (which you've already added to this PR)

PR #244, opened 2019-02-04
[QPPA-2731] Update 2019 quality measures and add measure specifications

kalvinwang:
Update 2019 quality measures including adding measure specifications for the first time in 2019 and updating some isToppedOutByProgram values.

Reviewer: @lamroger 

Tests: still passes unit tests

Jira: https://jira.cms.gov/browse/QPPA-2731


measures/2019/measures-data.json
```diff
@@ -242,7 +266,7 @@
     "isHighPriority": false,
     "isInverse": false,
     "isIcdImpacted": false,
-    "isToppedOutByProgram": false,
+    "isToppedOutByProgram": true,
```
kalvinwang: @sarahnw just a heads up that a bunch of isToppedOutByProgram values are changing, presumably intended

sarahnw: yep! we didn't have that info until we ran benchmarks so we have it now :)

PR #248, opened 2019-02-19
[QPPA-2832] Make 2018 QCDR measures uppercase

kalvinwang:
Currently, scoring in prod for 2018 measures is breaking because capitalization for 2018 QCDR measures in the repo doesn't match their capitalization in benchmarks. This PR makes QCDR measures uppercase, which is consistent with all other measures in all years. 

As a bonus it removes some unwanted `Õ` characters. This will also allow us to remove a bit of [special-case code](https://github.com/CMSgov/qpp-submissions-api/blob/18667ca1e0a6b96530ab10efcec49c38581ecc4f/tools/create-benchmarks/year3/calculate-deciles-lib.js#L655-L655&undercov=1) in the benchmarks script calculate-deciles-lib.js in the Submissions API.

Jira: https://jira.cms.gov/browse/QPPA-2832

Tests: still passes all unit tests

Reviewer: @sheldon-b


kalvinwang:
Sure, linked! Yes the deciles tools will continue to work as is.

PR #249, opened 2019-02-25
QPPA-2737: Include Updated IA Measures 

shebryant:
When reviewing a previous PR that updated IA measures, Kalvin noticed an inaccuracy in the description for measureId IA_BE_1 in the csv file. I uploaded an updated file that was provided and bumped the version, but didn't actually run the build script to update the measures json before merging. I have since run build and this PR includes the updates for that measure in the measures json file. Once approved, I'll merge and publish.

### DIFF
--- a/measures/2019/measures-data.json
+++ b/measures/2019/measures-data.json
@@ -10050,7 +10050,7 @@
   {
     "category": "ia",
     "title": "Use of certified EHR to capture patient reported outcomes",
-    "description": "use, etc.) or patient activation measures through use of certified EHR technology, containing this data in a separate queue for clinician recognition and review.",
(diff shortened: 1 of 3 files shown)


kalvinwang:
Thanks for noticing and fixing! 

1) Want to include a patch version bump even though it probably won't matter much in practice? If we just silently fix it then it could be confusing for someone expecting this fix in the previous release where it was referenced.

2) I think this change should (additionally) be made in the ia-measures.csv file? I thought the build process used that as the source-- did you omit it when running git add, or did something else happen?

shebryant:
@kalvinwang Regarding 1) I haven't published or released 1.16.2 yet. I'd intended to get this change merged and then publish and release. I'm happy to have 2 separate releases if you think that is best.

Regarding 2) the change to the csv file was taken care of when I merged the previous PR. Master currently includes the correct description.

kalvinwang:
Oh gotcha, never mind either one then. lgtm!

PR #256, opened 2019-04-01
QPPA-2824: Add 2019 PI Measure Exclusions

shebryant:
This PR adds 2019 PI measure exclusions, but also adds exclusion as an expected column name so that PI measure exclusions don't have to be hard coded in the future. Spoke with Sarah and she plans to request that the csv files we receive for measures data henceforth include exclusions.

Testing
* [x] Verified measures were properly generated locally

* https://jira.cms.gov/browse/QPPA-2824


sheldon-b:
Where is the exclusion field used? Will that be in the submissions api as part of a separate ticket?

shebryant:
Logic for using exclusions already exists in the submissions api. It was added last year.

PR #258, opened 2019-04-04
QPPSF-4331 - Fixed Clinical Cluster Exception for missing 2019 clusters

shadford:
Fixed Clinical Cluster Exception for missing 2019 clusters

#### Motivation for change

Currently receiving error for 2019 Clinical Cluster which will be added later. Discussed with @bijujoseph we decided it would be best that the measures data return an empty array instead of throwing an exception for years that are not found. 

#### What is being changed

Modified index.js... exports.getClinicalClusterData(performanceYear = 2017) to return an empty array for years not found.

#### Release checklist:

* [ ] [Package version updated](../CONTRIBUTING.md#versioning-publishing-and-creating-new-releases)
* [ ] Documentation updated

* https://jira.cms.gov/browse/QPPSF-4331

Reviewers:

@kalvinwang @KyleApfel @bijujoseph @sheldon-b @kyeah @shebryant

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "qpp-measures-data",
-  "version": "1.18.0",
+  "version": "1.18.1",
   "description": "Quality Payment Program Measures Data Repository",
   "repository": {
     "type": "git",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "qpp-measures-data",
-  "version": "1.18.1",
+  "version": "1.18.2",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "qpp-measures-data",
-  "version": "1.18.0",
+  "version": "1.18.1",
```
sheldon-b: please use `npm version patch` so that both package.json and package-lock.json get updated with the new version

shadford: ahh thanks!

PR #272, opened 2019-07-08
QPPA-3086: Update Measure Types

shebryant:
#### Motivation for change

Beginning with 2019 quality measures, we need to update the current list of supported measureTypes.

#### What is being changed
This PR removes logic that combines some measureTypes into other types, and instead keeps the currently reported measureType field in csv for quality measures 1 to 1 in their provided descriptions. 

For 2019, the supported measureTypes for quality measures should be:
      - efficiency

      - intermediateOutcome

      - outcome

      - patientEngagementExperience

      - process

      - structure

      - patientReportedOutcome

#### Release checklist:

* [x] Documentation updated

* https://jira.cms.gov/browse/QPPA-3086


scripts/measures/2019/import-quality-measures.js
```diff
@@ -126,10 +126,7 @@ const MEASURE_TYPES = {
   'efficiency': 'efficiency',
   'intermediate outcome': 'intermediateOutcome',
   'structure': 'structure',
-  'patient reported outcome': 'outcome',
-  'composite': 'outcome',
-  'cost/resource use': 'efficiency',
-  'clinical process effectiveness': 'process'
```
harrisj: These other columns no longer apply too?

shebryant: No, I don't believe so. I double checked with Sarah to confirm that the 7 measure types that remain are the only ones that should be present.

KyleApfel:
@shebryant Measure ACO321 is all lowercase. Measure names should be in caps.
https://github.com/CMSgov/qpp-measures-data/pull/272/files#diff-ffff7d0000ef3a45d0295ad56a50f2d2R8682

PR #287, opened 2019-09-03
QPPA-3374 update 2019 qcdr measures

sheldon-b:
#### Motivation for change

We were given a new CSV file for 2019 QCDR measures. This PR incorporates that new file.

#### What is being changed

The supplied CSV file had some strange encoding characters in it so I converted it to utf-8 using a text editor.

After that, I overwrote the existing `util/measures/2019/qcdr-measures.csv` file.

Finally, I ran the following command: `npm run build:measures 2019`

#### Release checklist:

* [ ] Documentation updated

* https://jira.cms.gov/browse/QPPA-3374


measures/2019/measures-data.json
```diff
@@ -57511,15 +57515,15 @@
...
       },
       {
         "name": "counseltobacco",
-        "description": "Risk Assessment or Counseling for Tobacco Use"
+        "description": "Risk Assessment or Counseling for Tobacco Use,"
       },
       {
         "name": "counselsubstance",
-        "description": "Risk Assessment or Counseling for Other Substance Use"
+        "description": "Risk Assessment or Counseling for Other Substance Use, and"
```
sheldon-b: This also seems unusual @sarahnw

measures/2019/measures-data.json
```diff
@@ -46928,7 +46928,7 @@
...
     "nqfId": null,
     "nationalQualityStrategyDomain": "Communication and Care Coordination",
     "measureType": "structure",
```
shebryant: This looks a bit strange, `nBB-COC-PC=(∑_1^k _((BB-COC)(n_k ))_)`, think it's what is intended?

measures/2019/measures-data.json
```diff
@@ -47936,7 +47936,7 @@
     "strata": [
       {
         "name": "delayed",
-        "description": "Advance 3B: Procedure delayed from scheduled start time"
+        "description": "Advance 3B: Procedure delayed from scheduled start time;"
       },
```
shebryant: The semicolon, weird, right? Also intended?

measures/2019/measures-data.json
shebryant: I'd agree that this looks a bit strange.

measures/2019/measures-data.json
sheldon-b: I think so, looks like a sigma notation math formula

measures/2019/measures-data.json
sheldon-b: I doubt it's intended but I feel like it should be fixed by whoever gives us the csv instead of by us. What do you think @sarahnw ?

measures/2019/measures-data.json
sarahnw: I think the \n\n is the paragraph breaks. also this comment is on the old version so not sure what specifically I'm supposed to review? the green side looks like the right formula except for the \n\n but I'm not sure how to get rid of those except to take out all of the extra lines in the excel?

measures/2019/measures-data.json
sarahnw: I don't think it matters. QCDR measures aren't shown in the shopping cart. These are how the measures are delivered. If we think it's that important we can give them editing feedback b/c it's displayed potentially elsewhere, but in the grand scheme of things I don't think it matters.

measures/2019/measures-data.json
sarahnw: it's how it's formatted on the spreadsheet.

PR #291, opened 2019-10-17
Fix measure 370

saquino0827:
https://jira.cms.gov/browse/QPPA-3428
https://jira.cms.gov/browse/QPPSF-5272

#### Motivation for change

This is to fix an issue where the data was being set to `singlePerformanceRate` for measure 370 when extracting data from the EP_EC_eCQM_2018-05.zip. This zip is the 2019 version of the eCQMs uuid's. Also, fixing the strata names for measure 370.

#### What is being changed

Measure 370's metric type is being changed from `singlePerformanceRate` to `multiPerformanceRate`. Strata names have been added for measure 370.

#### Release checklist:

* [x] Documentation updated


saquino0827:
@shebryant Here is a PR to fix the script to make measure 370 into a `multiPerformanceRate`. The script currently sets any measure with multiple strata as a multiPerformanceRate which is what caused measure 370 to be invalid.

saquino0827:
There is another issue I will be adding to this PR related to this measure. Please hold off reviews until I have linked the new ticket and completed it.

saquino0827:
Description edited to add new task and comment on what was fixed.

util/measures/2019/generated-ecqm-data.json
```diff
@@ -451,7 +483,7 @@
         }
       }
     ],
-    "metricType": "singlePerformanceRate"
```
branonbarrett: Does this file not need the same strata as the `measures-data.json`? Would adding that keep you from having to do the conditional check and pushing in the `{ name: '18+' }` strata in the`nrich-measures-data.js` code?

saquino0827: This file is generated from a separate script that takes in a zip file with measures. The measures themselves do not have a separate `12-17` and `18+` strata. There is another file at `util/measure/2019/quality-strata.csv` with strata names to associate with measure id's. But it associates them so by linking them with a numerator UUID. Since the IG doesn't actually have a numerator UUID for 18+ I manually pushed the name `18+` once it gets to the correct measure. I could see if I could take the name from the file that passes in that strata name if the manual push is a concern.

branonbarrett: I'm mostly just curious and since that file is 2019 specific I don't see any real problem with just pushing into the array

PR #293, opened 2019-10-24
Remove measure 226 benchmarks

saquino0827:
https://jira.cms.gov/browse/QPPSF-5268

#### Motivation for change

Measure 226 needs to be suppressed on the web interface side.

#### What is being changed

Measure 226 benchmarks are being removed.

#### Release checklist:
Tasks that must be done prior to merging this PR, including testing.

* [x] [Package version updated](https://github.com/CMSgov/qpp-measures-data/blob/master/CONTRIBUTING.md#versioning-publishing-and-creating-new-releases)
* [x] Documentation updated
* [x] Unit tests added/passing
* [x] Verified working locally

PR #294, opened 2019-11-04
Bump version to 1.31.4

sheldon-b:
#### Motivation for change

Both #291 and #293 bumped the version to 1.31.3 but only one of those changes were included when 1.31.3 was published to npm. This resulted in only one of the features being released into PROD when we were expecting both of them to go out.

#### What is being changed

Bumps the qpp-measures-data repo to 1.31.4 so we can publish the measure 370 change without overriding the version of 1.31.3 that is already deployed to PROD.

#### Future work

The ticket QPPA-3525 is for a feature enabler to add CI to qpp-measures-data. That CI would have helped avoid this scenario and others like it but allowing us to utilize automation for more quality checks and process steps.

#### Release checklist:

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "qpp-measures-data",
-  "version": "1.31.3",
+  "version": "1.31.4",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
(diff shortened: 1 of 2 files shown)


sheldon-b:
This has been published to npmjs

PR #296, opened 2019-11-06
QPPSF-4894 - Calculate PY2020 Historic Quality Benchmarks

shadford:
Needed to create 2020 measures in order to generate 2020 benchmarks.

#### Motivation for change

Calculate PY2020 Historic Quality Benchmarks

#### What is being changed

Calculate PY2020 Historic Quality Benchmarks

#### Release checklist:

* [ ] Documentation updated

* https://jira.cms.gov/browse/QPPSF-4894

@harrisj @lamroger @sheldon-b @kyeah @shebryant @KyleApfel @NateTheGreatt @Jpec07 @bijujoseph


sarahnw:
@shadford are these the actual 2020 measures? I have files for them but have been waiting for some confirmation before publishing them. Please let me know where these come from. This is public so I want to make sure we know what we're publishing here so we can inform our external users appropriately.

shadford:
Hi Sarah,

There are measures based off the benchmarks we published, we'll need to
update them once they are ready. I'm not exactly sure why we are publishing
the benchmarks before the measures data but the benchmark ticket was high
priority for some reason.

sarahnw:
thanks @shadford. I know that benchmarks are important to get out there, but they would be dependent on the actual list of 2020 measures, both for CQMs and QCDR. do you know where you got these from? We've got tickets to add the actual quality, IA, and PI measures and I'm not sure if these are right or not. Considering they have 2019 measure specs still attached, these could just be the 2019 measures and I'm not sure if it is appropriate to be publishing them in the 2020 folder as this is considered the source of truth for QPP measures. benchmarks are necessary to publish, but there should be more comms between teams so we can sync measures and info to the registries. plus these will need to be published in the resource library and I'm not aware of where we stand on that clearance.

shadford:
Yes, can you sync up with Kyle Hoy? I'll remove the measures for now, and
tag a note that the benchmarks will not be regenerated until the measure
have been received.

sarahnw:
Steven says the benchmarks can be published, but if we can not publish the measures file then that would be helpful. We've got those real ones coming that'll impact benchmarks too. I don't think either should be published without the other, but benchmarks we can update once the final 2020 measures is ready. According to Steven these are the 2019 measures and I'd rather not have registries running off with these thinking they're the final 2020 ones.

shadford:
Ok, sounds good. I'll whip up a PR to remove them right away.

sarahnw:
Thanks!! sorry for the crazy. I have the real 2020 measures that I'm
reviewing the file on tonight. I have a question out to PIMMS on them
still, but I'm hoping we'll get them up this week.

shadford:
I should have checked first, I thought it was a little weird request
without the 2020 measures being there yet. Thanks for letting me know

PR #300, opened 2019-11-22
QPPA-3293, QPPA-3294, QPPA-3295 - Add 2020 Quality, PI, IA measures

chouinar:
https://jira.cms.gov/browse/QPPA-3293
https://jira.cms.gov/browse/QPPA-3294
https://jira.cms.gov/browse/QPPA-3295

#### What is being changed

This adds the 2020 measures for quality, PI and IA measures as well as makes a few adjustments to the scripts for consuming 2020 data, although these are largely the same as 2019s ingestion scripts. Also includes some minor updates for 2019 PI measures.

Detailed information about what is being changed and rationale for interesting decisions made

#### Release checklist:

* [x] Documentation updated


util/measures/2019/pi-measures.csv
```diff
@@ -1,40 +1,40 @@
-measureId,title,description,isRequired,metricType,firstPerformanceYear,weight,objective,isBonus,reportingCategory,substitutes,measureSpecification,exclusion
```
sheldon-b: Why was this file changed?

sheldon-b:
Oof, that's kind of a negative tone. Thank you for doing this 🙏 just curious about what sort of qa process we have in place, if any

util/measures/2019/pi-measures.csv
chouinar: There was an update for 2019 data as well, mostly just affecting a few descriptions.

chouinar:
@sheldon-b  Scripts were largely unchanged from 2019. A few column names changed, a few new valid values existed in some fields. Some columns came in as something like "My Value" when they needed to be camelCased as "myValue" so I camel-case the column.

If it helps, I could do some sort of revert, copy the scripts from 2019->2020, and then put my changed versions in a separate commit so you can see the actual changes. Let me know if that would help you review.

As for tests, let me look if I should add some more. There is a test that checks that the 2020 measures is valid that is based on the config values, so it wasn't explicitly written.

sheldon-b:
Okay, that's fine, thank you. I'm satisfied that _some_ tests are running. We can think more about tests in the future

PR #301, opened 2019-11-26
QPPA-3561, QPPA-3562 - Update 2019 QCDR and quality measures

chouinar:
https://jira.cms.gov/browse/QPPA-3561
https://jira.cms.gov/browse/QPPA-3562

#### What is being changed
QPPA-3561 -Removes ARCO12 and ARCO13 and changes PIMSH5 to an inverse
QPPA-3562 - This will update measure spec links for eCQMs and fix two characters in the strata for measure 394.

Measure 326 and 392 both changed to `icdImpacted = true` as well.

#### Release checklist:

* [x] [Package version updated](https://github.com/CMSgov/qpp-measures-data/blob/master/CONTRIBUTING.md#versioning-publishing-and-creating-new-releases) - will update once clarification on the above question has been determined (Waiting on Sarah who is on PTO until next week)


measures/2019/measures-data.json
```diff
@@ -31797,7 +31797,7 @@
         "description": "Overall Percentage for patients (aged 5-50 years) with well-controlled asthma, without elevated risk of exacerbation"
       },
       {
-        "name": "5-17",
+        "name": "17-May",
```
jenjaina: This doesn't look like it was supposed to happen

chouinar: Good catch, seems it was an issue with excel trying to read ranges as dates and 'conveniently' converting it.

sarahnw:
@chouinar did you see my comment on QPPA-3561? I don't think the qcdr measure script is correctly identifying high priority measures. I don't see those flags correctly marked in here either. let me know if you want me to make a new ticket.

chouinar:
@sarahnw I had not seen that, let me check what's going on.

PR #305, opened 2019-12-04
QPPA-3532: Delete 2019 Benchmarks for Measures that No Longer Exist

shebryant:
#### Motivation for change

Get /api/submissions/public/benchmarks?measureId=AAD3 returns a 422 error with the message "Invalid value: AAD3 for filter parameter: measureId." Yet this entry can be found in the 2019 benchmark file: https://raw.githubusercontent.com/CMSgov/qpp-measures-data/master/benchmarks/2019.json. The error occurs for a total of 26 measure ids.

#### What is being changed

After looking at the measureIds for which the error is occurring, it was determined that the specific ids are no longer valid ids for Y3. Therefore, data for measures with these ids was deleted from `benchmarks.csv` and the script to build benchmarks was run again, resulting in the removal of benchmarks for these ids in the 2019 benchmark JSON. Ultimately, the 422 error is and should be returned, because the measureIds are invalid.

#### Release checklist:

* [ ] Documentation updated

* https://jira.cms.gov/browse/QPPA-3532

### DIFF
--- a/benchmarks/2019.json
+++ b/benchmarks/2019.json
@@ -4554,24 +4554,6 @@
     ]
   },
   {
-    "measureId": "AAD1",
-    "benchmarkYear": 2017,
-    "performanceYear": 2019,
-    "submissionMethod": "registry",
-    "isToppedOut": false,
-    "deciles": [
-      0,
-      10.16,
-      18.75,
-      25.13,
-      35.84,
-      43.48,
-      57.08,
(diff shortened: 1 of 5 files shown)


chouinar:
Hmm, if we do this and then I merge #310 I think I'd just end up reverting the change made here. I think the work from these two need to be joined together.

PR #306, opened 2019-12-05
QPPA-3430 Update 2019 schema to error on extra fields

sheldon-b:
#### Motivation for change

Update 2019 and 2020 measure schemas to be strict about "additional properties". This means that if a measure has properties not in the schema it will fail. The current behavior is to allow those additional properties which is problematic for A&R (and is just a great recipe for bugs throughout the system).

#### What is being changed

Makes use of the `additionalProperties: false` json schema feature to validate that there are no additional properties. In order to make this work I had to get rid of the use of `anyOf` in the `measure` schema because `additionalProperties: false` only validates things at the top level schema where it is used. Instead I've made use of the `$merge` feature to merge in properties which does work with `additionalProperties: false`.

Also refactors the `scripts/validate-data.js` tool so we can use it in tests without having to execute shell commands.

#### Release checklist:

* https://jira.cms.gov/browse/QPPA-3430


measures/2020/measures-schema.yaml
```diff
@@ -1,11 +1,21 @@
...
 $schema: http://json-schema.org/draft-07/schema
 type: array
 items: { $ref: #/definitions/measure }
 uniqueItemProperties: ['measureId'] # used by ajv-keywords
 
 definitions:
   measure:
+    title: 'Measure'
+    type: object
+    anyOf:
```
harrisj: I wonder if this should be an `anyOf` or a `oneOf`?

sheldon-b: In practice it shouldn't be possible for an object to match multiple of these schemas since each one has `additionalProperties: false` and each one also has a different const value of `category`. `anyOf` will be faster since it will stop validating once the data matches any of them. Since this schema is only used by unit tests, though, speed isn't really a concern at all. No strong feeling either way for me

harrisj: Sounds good then

harrisj:
Let me tag in Jenni Kissinger to review

measures/2019/measures-schema.yaml
```diff
@@ -30,158 +37,183 @@ definitions:
       measureSpecification:
         description: URL link for Measure Specification PDF to download by Submission Method.
         items: { $ref: #/definitions/measureSpecification }
+      isClinicalGuidelineChanged:
```
jenjaina: I thought I only saw this on quality measures, is it true that it applies to all measures?

sheldon-b: Ah, yes, you're right. Good catch. Fixed in e287e6e.

PR #283, opened 2019-08-14
QPPA-3341: Replace benchmark.csv with newly generated from submissions api

lamroger:
#### Motivation for change

Measurements were being double counted when generating historical benchmarks when different submitters were submitting for the same performance year, measure, and submission method in the same submission. We fixed this in https://github.com/CMSgov/qpp-submissions-api/commit/c28268277435a39b06556390f07548d0381a26f4#diff-e34e0d0d950a23e329c97e3401542e73 and regenerated the historical benchmarks. 

Note: The final scoring team will be responsible for generating historical benchmarks next year.

#### What is being changed

The new historical benchmarks were generated and added in this commit - https://github.com/CMSgov/qpp-measures-data/pull/283/commits/61377a985ca91f7e04aa61ab6b4f503dc80a5005

Between the time the first time this "base" historical benchmarks were generated, there were two notable changes to the data. 
https://github.com/CMSgov/qpp-measures-data/commits/master/benchmarks/2019.json

1. CAHPS benchmarks were added. https://github.com/CMSgov/qpp-measures-data/commit/e4b84865ed231bda553925faf334c5545ac77904#diff-19cc2b2b0a838f96141210f2b77688c9
2. Web Interface benchmarks were added. https://github.com/CMSgov/qpp-measures-data/commit/e8a7e2912f9605e57437fdde02559538f4274f1c#diff-19cc2b2b0a838f96141210f2b77688c9

Since the CAHPS benchmarks were added in another "layer", they did not need to be changed in the base layer.
Web Interface benchmarks did need to be modified in the base layer. https://github.com/CMSgov/qpp-measures-data/pull/283/commits/6a9260d2e635565d06b4e073f100eda4d3150c37

Lastly, some benchmarks were removed in the original base layer. https://github.com/CMSgov/qpp-measures-data/pull/234 
```
[...]

PR #310, opened 2019-12-16
QPPA-3580, QPPA-3536 - Update benchmarks for isToppedOutByProgram, revert decile change, and delete benchmarks for measures that no longer exist

chouinar:
https://jira.cms.gov/browse/QPPA-3580
https://jira.cms.gov/browse/QPPA-3536

#### What is being changed

This adds an isToppedOutByProgram field to various benchmarks which is the new 'SevenPointCap' column in the input file.

Additionally, the submission method is now determined using the collection type column. The mapping of new names back to the old is from https://confluence.cms.gov/pages/viewpage.action?spaceKey=QPPARCH&title=Task%3A+Aggregate+Multiple+Submissions+and+Update+Terminology
We are sticking with the older naming schema as the API expects that still.

This also reverts the benchmark changes from #283 as the program has decided to not update these values.

#### Release checklist:

* [ ] Documentation updated

### DIFF
--- a/scripts/benchmarks/format-benchmark-record.js
+++ b/scripts/benchmarks/format-benchmark-record.js
@@ -13,14 +13,10 @@ const Constants = require('../../constants.js');
  * @enum {string}
  */
 const SUBMISSION_METHOD_MAP = {
-  'claims': 'claims',
-  'registry': 'registry',
-  'registry/qcdr': 'registry',
-  'cmswebinterface': 'cmsWebInterface',
-  'administrativeclaims': 'administrativeClaims',
-  'ehr': 'electronicHealthRecord',
-  'cmsapprovedcahpsvendor': 'certifiedSurveyVendor',
-  'cahpssurveyvendor': 'certifiedSurveyVendor'
+  'ecqm': 'electronicHealthRecord',
+  'medicarepartbclaims': 'claims',
+  'mipscqm': 'registry',
+  'qcdrmeasure': 'registry'
 };
 
 /**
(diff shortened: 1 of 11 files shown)


jenjaina:
This field should be added to the `2019/benchmarks-schema.yaml` file.

scripts/benchmarks/format-benchmark-record.js
```diff
@@ -13,14 +13,10 @@ const Constants = require('../../constants.js');
...
-  'registry/qcdr': 'registry',
-  'cmswebinterface': 'cmsWebInterface',
-  'administrativeclaims': 'administrativeClaims',
-  'ehr': 'electronicHealthRecord',
-  'cmsapprovedcahpsvendor': 'certifiedSurveyVendor',
-  'cahpssurveyvendor': 'certifiedSurveyVendor'
+  'ecqm': 'electronicHealthRecord',
+  'medicarepartbclaims': 'claims',
+  'mipscqm': 'registry',
+  'qcdrmeasure': 'registry'
```
harrisj: I suppose this is fine, but I wanted to check

chouinar: All of them got renamed, see https://confluence.cms.gov/pages/viewpage.action?spaceKey=QPPARCH&title=Task%3A+Aggregate+Multiple+Submissions+and+Update+Terminology

Beyond that, it removes spaces + lower-cases everything so the keys look a bit weird. Kept the logic alone for now and just had the weird keys.

scripts/benchmarks/format-benchmark-record.js
```diff
@@ -13,14 +13,10 @@ const Constants = require('../../constants.js');
...
-  'claims': 'claims',
-  'registry': 'registry',
-  'registry/qcdr': 'registry',
-  'cmswebinterface': 'cmsWebInterface',
-  'administrativeclaims': 'administrativeClaims',
-  'ehr': 'electronicHealthRecord',
-  'cmsapprovedcahpsvendor': 'certifiedSurveyVendor',
-  'cahpssurveyvendor': 'certifiedSurveyVendor'
+  'ecqm': 'electronicHealthRecord',
+  'medicarepartbclaims': 'claims',
```
jheltonsb: does this effect my team? We used submission method "claims"

chouinar: In the API, it will unaffected. This effectively maps from the new names back to the old submission method names which the API will continue to use.

jheltonsb: Perfect. thank you!

PR #315, opened 2020-01-17
QPPA-3665: Update measure specification with CQM, PI, WI, and Claims links

bartcone:
#### Motivation for change

Measure specification links have been provided for CQM, PI, WI, and Claims for PY2020.

#### What is being changed

`measureSpecification:` has been updated with the appropriate link for a given measure.

`measureSpecification.default` => `PI`
`measureSpecification.registry` => `CQM`
`measureSpecification.cmsWebInterface` => `WI`
`measureSpecification.claims` => `Claims`

#### Release checklist:
Tasks that must be done prior to merging this PR, including testing.

* [ ] [Package version updated](https://github.com/CMSgov/qpp-measures-data/blob/master/CONTRIBUTING.md#versioning-publishing-and-creating-new-releases)
* [ ] Documentation updated
* [x] Unit tests added/passing
* [x] Verified working locally

##### Associated JIRA tickets:
* https://jira.cms.gov/browse/QPPA-3665

PR #316, opened 2020-01-17
QPPA-3668: Added cost measures for PY2020

bartcone:
#### Motivation for change

Cost measures were provided for PY2020.

#### What is being changed

Cost measures have been added to the overall `measures-data.json`. Updated the 2020 `measures-schema.yaml` to allow for `null` values in `overallAlgorithm` property. `build-2020-measures` run script now updated to process cost measures.

#### Release checklist:

* [ ] Documentation updated

* https://jira.cms.gov/browse/QPPA-3668


measures/2020/measures-data.json
```diff
@@ -9517,5 +9517,345 @@
...
+    "metricType": "costScore",
+    "firstPerformanceYear": 2017,
+    "lastPerformanceYear": null,
+    "isInverse": true,
+    "overallAlgorithm": "simpleAverage",
+    "submissionMethods": [
+      "administrativeClaims"
+    ],
+    "measureSpecification": {
+      "measureInformation": "https://qpp.cms.gov/docs/cost_specifications/2019-12-17-mif-tpcc.pdf"
```
joshboley: One change.  We determined we were going to use `default`, not `measureInformation`.

bartcone: For cost measures? We used `default` for `PI` in #315.

joshboley: Yes.  As a pattern, if the measure spec is not associated with a `submissionMethod` (which `measureInformation` is not), then the key will be `default`

bartcone: @joshboley addressed feedback.

PR #320, opened 2020-01-27
QPPA-3758: Updated strata order for 226 measure

bartcone:
#### Motivation for change

2020 226 measureId had the incorrect strata ordering

#### What is being changed

strata order for 226 measureId

#### Release checklist:

* https://jira.cms.gov/browse/QPPA-3758


bartcone:
Bundling with `1.40.0` release.

joshboley:
@branonbarrett no idea, but Steven said it mattered

bartcone:
> Why does the order matter? If the order is truly driving some piece of the SE code we should add a field to identify which strata is used for performance to make it explicit.

@branonbarrett i agree - do we want to create a ticket to track this?

PR #327, opened 2020-03-16
QPPA-3960, QPPSF-5794: Measure spec updates

bartcone:
#### Motivation for change

Measures 283 and 383 were updated with the correct links as part of QPPA-3960. This PR also includes measure spec link updates for PY 2017 and 2018. PY 2017 source file was updated and the generated file was manually updated as well. A full regeneration of 2017 measures was NOT run.

#### What is being changed

Measure specification links for given measures.

#### Release checklist:

* [x] Documentation updated

* https://jira.cms.gov/browse/QPPA-3960
* https://jira.cms.gov/browse/QPPSF-5794


bartcone:
It looks like `util/measures/2018/quality-measures.csv` was previously one single line, but after updating a few of measure spec links it added new line chars.

bartcone:
> * 1 nit. Not sure why we need to wait 5 minutes for network glitch.
> * Do we need/want to add the "freeze" in the generation script so 2017 doesn't get overwritten?
> 
> Other than that LGTM

@branonbarrett i upped that mainly because it's going through all PYs now and hitting the network for each link and it was timing out. It could probably be lowered some but since it's only run once a day on a cron, I didn't think it was a big deal.

As far as the 2017 stuff goes, there's not a generation script for it to begin with so I think we're fine. You'd have to go pretty far out of your way to overwrite 2017 data.

PR #340, opened 2020-05-21
Updated 2019 performance benchmarks

branonbarrett:
#### Motivation for change
Publishing performance year benchmarks from FS 2019 run 527.

#### What is being changed
- Added 2019 performance-benchmarks
- Ran script to update `benchmarks/2019.json`
- Updated version to 3.0.0. Why go from 1.45.0 to 3.0.0? I feel like the major version number should correspond to the program year. Since this is the close of Y3 then 3.0.0 is fitting. Measure updates for Y4 should increment to 4.x.x.

#### Release checklist:
* [x] Documentation updated


benchmarks/2019.json
```diff
@@ -4622,14 +4850,14 @@
     "benchmarkYear": 2019,
     "submissionMethod": "registry",
     "deciles": [
-      96.4286,
+      91.0237,
       62.2222,
-      53.8961,
```
KyleApfel, line 4627: In this case, the performance benchmark is overriding the historical benchmark which is *not* okay. In the case a historical benchmark exists, the performance benchmark is not applied. @branonbarrett

branonbarrett: `"benchmarkYear": 2019,` indicates this is not a historical benchmark right?

PR #346, opened 2020-06-09
QPPA-4193 Fix Cost Measure Spec Links

pgrocholski:
#### What is being changed

Fixed Cost Measure Spec links

#### Release checklist:

* https://jira.cms.gov/browse/QPPA-4193


joshboley:
@pgrocholski I don't see the final json file updated.  Only the staging one.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "qpp-measures-data",
-  "version": "3.0.1",
+  "version": "3.0.2",
```
branonbarrett, line 3: Don't need to change this with this PR. Should change the base to develop so we can PR to a release branch from there. We'll update the version on that PR

PR #349, opened 2020-06-10
QPPA-3756 Github Action Workflow

pgrocholski:
#### Motivation for change

Github workflow for automating release management and notifications.
[Tech Spec](https://confluence.cms.gov/display/QPPARCH/Tech+Spec%3A+Automated+Publication+of+New+Measures+and+updates)

#### What is being changed
* test runs for both feature and release PRs
    - additional version bump check for release PRs
* draft of release page on release merge to master
* on prerelease:
  - publish to gpr
  - publish to npm
  - dispatch of event to dependent projects
  - notification of relevant slack channels of publish

#### Configuration
The following secrets need to be set by an admin:
* `NPM_TOKEN` for publishing to npm
* `GITHUB_USER_TOKEN` for dispatching of events to other repos
* `SLACK_WEBHOOK` for dispatching notifications to slack

#### Notes
The slack webhook needs to be created by someone with permissions, and the integration will likely require additional configuration/testing. Additional build steps such as actually building the json could be done in the feature branch pr action should we want to do that/are ok with the risks associated with it.

* https://jira.cms.gov/browse/QPPA-3756

### DIFF
--- a/.github/workflows/pr-feature.yml
+++ b/.github/workflows/pr-feature.yml
@@ -9,9 +9,13 @@ jobs:
   build:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v2
-      - uses: actions/setup-node@v1
+      - name: Checkout Codebase
+        uses: actions/checkout@v2
+      - name: Configure Node version and registry
+        uses: actions/setup-node@v1
         with:
           node-version: 12
-      - run: npm ci
-      - run: npm test
+      - name: Install Dependencies
+        run: npm ci
+      - name: Run Tests
+        run: npm test
--- a/.github/workflows/pr-feature.yml
+++ b/.github/workflows/pr-feature.yml
@@ -1,0 +1,17 @@
+name: QPP Measures Data Feature PR CI/Checks
+on: 
+  pull_request:
+    types: [opened, synchronize]
+    branches: [ develop ]
+
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v1
+        with:
+          node-version: 12
+      - run: npm ci
+      - run: npm test
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -60,9 +60,9 @@ jobs:
           - CMSgov/qpp-submissions-api
     steps:
       - name: Repository Dispatch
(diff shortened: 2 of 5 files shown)


.github/workflows/publish.yml
```diff
@@ -0,0 +1,89 @@
+name: QPP Measures Data Publish
+
+on:
+  release:
+    types: [ prereleased ]
+
+jobs:
+
+  publish-gpr:
```
bijujoseph, line 9: we can remove this pice later (this is pending approval)

.github/workflows/publish.yml
```diff
@@ -0,0 +1,89 @@
...
+    runs-on: ubuntu-latest
+    needs: publish-npm
+    strategy:
+      matrix:
+        channel:
+          - p-qppsf-scoring
+          - p-wi
+          - p-qppsf-api-team
+    steps:
+    - uses: actions/checkout@v2
```
bijujoseph, line 81: why do we need checkout here ? 
Also in every file, use the documentation style (see the `name`  filed) 
```

      - name: Checkout Codebase
        uses: actions/checkout@v2

      - name: Configure Node version and registry
        uses: actions/setup-node@v1
        with:
          node-version: 12
          
```

.github/workflows/pr-feature.yml
```diff
@@ -0,0 +1,17 @@
+name: QPP Measures Data Feature PR CI/Checks
+on: 
+  pull_request:
+    types: [opened, synchronize]
+    branches: [ develop ]
```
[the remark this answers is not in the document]
branonbarrett: We had a case just two days ago that begs to differ. If multiple changes are in flight we may not want to publish each one individually with a merge to master. Also, like the case on Tuesday, we may need to get a set of changes out before other changes that have not been properly PRed yet.

.github/workflows/publish.yml
```diff
@@ -0,0 +1,89 @@
...
+  # make an announcement to relevant slack channels
+  notify-slack:
+    runs-on: ubuntu-latest
+    needs: publish-npm
+    strategy:
+      matrix:
+        channel:
+          - p-qppsf-scoring
+          - p-wi
+          - p-qppsf-api-team
```
joshboley, line 79: We should add the UI channel here.  We can't update their repo since it's in enterprise github, but we could at least let them know in slack.  FE team, we should be able to notify on their repo _and_ slack.

PR #350, opened 2020-06-17
Gitflow backfill to develop

branonbarrett:
Gitflow


joshboley:
Is this a one-time thing since you just switched to git flow?  Seems like something we wouldn't normally want to do

branonbarrett:
> Is this a one-time thing since you just switched to git flow? Seems like something we wouldn't normally want to do

If the package version is updated on the release branch we will have to merge this back to develop from the release branch or master after the release

joshboley:
Ok, makes sense.  I guess in my mind, backfilling from the release branch seems "better", though I can't think of a great justification.

PR #365, opened 2020-08-04
QPPC-1168 py2020 measures update

mike-sanderson:
#### Motivation for change

Received updated Claims Single Source CSV file defining PY2020 scoring logic for claims data

#### What is being changed

The 2020 Claims Single Source CSV file was added to the repo. The eligibility and performance option data from it was integrated with the 2020 measures-data.json file by updating the build-2020-measures.sh script and rerunning it.

Also a comparison bug for merging claims eligibility options was found and fixed in the single_source_conversion_helpers.py file.

Unit tests were added to test the claims single source merge functionality as well as the bug fix in single_source_conversion_helpers.py

All unit tests pass. npm test coverage of new qpp-measures-data/scripts/measures/lib/merge-claims-related-data.js is 100%

#### Release checklist:

* [ ] Documentation updated

* https://jira.cms.gov/browse/QPPC-1168


claims-related/tests/test_single_source_conversion_helpers.py
```diff
@@ -140,6 +140,28 @@ def test_find_min_max_age_years():
...
+    expected_output = {
+        "01.00": {
+            "eligibilityOptions": ["elOpt1", "elOpt2", "elOpt3"],
+            "performanceOptions": [{"pOpt1": "pval1"}, {"pOpt2": "pval2"}, {"pOpt3": "pval3"}]
+        }
+    }
+    output = single_source_conversion_helpers.merge_multiple_eligibility_options(input)
+    assert output == expected_output
+
+
```
jared-baker, line 164: This was to test 418 and 418.01 correct?

claims-related/scripts/single_source_conversion_helpers.py
```diff
@@ -220,12 +221,15 @@ def merge_multiple_eligibility_options(single_source_dict):
...
-                    single_source_dict[measure]['performanceOptions']
+
+            new_performance_options = [json.loads(x) for x in (
+                    set(json.dumps(x_new, sort_keys=True)
+                        for x_new in single_source_dict[measure]['performanceOptions']) -
+                    set(json.dumps(x_old, sort_keys=True)
+                        for x_old in single_source_dict[measure_to_update]['performanceOptions'])
+            )]
+
+            single_source_dict[measure_to_update]['performanceOptions'] += new_performance_options
```
jared-baker, line 232: Removing duplicates from the old way we processed the file?

claims-related/tests/test_single_source_conversion_helpers.py
mike-sanderson: Any multipart measure should go through this logic, so this would include 226, 418, and 24. in generating the 2020 measure data with the updated single source definitions, 418 is the measure where the bug occurred that this test is designed to make sure doesn't happen again.

claims-related/scripts/single_source_conversion_helpers.py
mike-sanderson: Yes. The previous way the file processed checked for an exact match, and if it wasn't found, would combine the two measures. We already de-duplicate rows in the initial read, so the only way we would get duplicates is in these multipart measure definitions. It looks like the previous code expected an exact match or something totally different so it probably didn't consider a partial match. If we're uncomfortable with that level of deduplication we can change line 232 to be `single_source_dict[measure_to_update]['performanceOptions'] += single_source_dict[measure]['performanceOptions']`, but i believe the original intent here was to get the union of performance options.

branonbarrett:
> We need to sort out the versioning. I'm not sure if I need to put a commit on your branch for the version update before we merge or what

@jared-baker Once this is merged to develop and if this needs to be released (i.e. there are no other changes from other teams at this time) we need to cut a `release/v3.1.5` branch and bump the package version to 3.1.5

PR #399, opened 2020-12-14
PY2021 Measures for Benchmarks

pgrocholski:
Merge in the partial PY2021 measure data for benchmark generation. We will likely want to hold off on publishing a package until the full data set is compiled.


KyleApfel (approved):
Looks good. Do you know when we'll get the rest of the full data set compiled and does that include the measure specification's?

bartcone (approved):

PR #436, opened 2021-03-05
QPPA-4987: 2020-2021 CAHPS Measures

pgrocholski:
Load PY 2020 and 2021 measures for CAHPS reporting. Measures are unchanged from 2019. ACO scores will not be loaded for these years, but measures are included for the sake of completeness.

* https://jira.cms.gov/browse/QPPA-4987


michelingle (changes requested):
In PY2021 ACO_321 is deprecated.  We should not load the associated ACO SSM measures.

nicholas-gates (approved):

PR #464, opened 2021-08-16
QPPSF-8206: Fix benchmark year for 2020 WI

pkasarski:
#### Motivation for change

Non-performance benchmarks should use a benchmark year two years prior to the performance year.

#### What is being changed

#387 fixed the output of the benchmark script, but not the WI input. If you re-run the script, the output has 2020 for the benchmark year for WI measures. So the input `staging/2021/benchmarks/json/wi-benchmarks.json` file needs to be changed.

#### Release checklist:

* [ ] Documentation updated

* https://jira.cms.gov/browse/QPPA-8206

### DIFF
--- a/staging/2020/benchmarks/json/wi-benchmarks.json
+++ b/staging/2020/benchmarks/json/wi-benchmarks.json
@@ -1,7 +1,7 @@
 [
   {
     "measureId": "110",
-    "benchmarkYear": 2020,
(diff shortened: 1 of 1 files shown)


shadford (approved):
Looks Good! Thank you!

pgrocholski (approved):

PR #512, opened 2021-12-17
QPPA-5949 remove IROMS benchmarks 2021

ckawell:
Removed IROMS11 - 20 from 2021 benchmarks.

* https://jira.cms.gov/browse/QPPA-5949


staging/2021/benchmarks/source.csv
```diff
@@ -510,16 +510,16 @@ Improved Visual Acuity after Vitrectomy for Complications of Diabetic Retinopath
...
-"Failure to Progress (FTP): Proportion of patients failing to achieve a Minimal Clinically Important Difference (MCID) to indicate functional improvement in knee rehabilitation of patients with knee injury measured via their validated Knee Outcome Survey (KOS) score, or equivalent instrument which has undergone peer reviewed published validation and demonstrates a peer reviewed published MCID.",IROMS11,QCDR Measure,Patient Reported Outcome,Y,26.800022172949,Y,42.45 - 32.6,32.59 - 30.01,30.0 - 27.05,27.04 - 25.01,25.0 - 22.62,22.61 - 21.63,21.62 - 15.59,<= 15.58,N,N
+"Failure to Progress (FTP): Proportion of patients failing to achieve a Minimal Clinically Important Difference (MCID) to indicate functional improvement in knee rehabilitation of patients with knee injury measured via their validated Knee Outcome Survey (KOS) score, or equivalent instrument which has undergone peer reviewed published validation and demonstrates a peer reviewed published MCID.",IROMS11,QCDR Measure,Patient Reported Outcome,Y,--,N,--,--,--,--,--,--,--,--,N,N
```
nicholas-gates, line 513: Hey - Not sure if this is an issue or not but I noticed that the other files updated 10 measures and this updates 9. Either that our I'm miscounting, lol. Can you double-check why this file has 9 updates versus 10.

nicholas-gates (changes requested):

staging/2021/benchmarks/source.csv
ckawell: They're both updating Decile 3 through 10 and the Average, but the average is located in a different position for each csv. In the first, the average comes after the IncludeInBenchmarks field (The Y that was changed to N), but in the second file it comes before the N.

nicholas-gates (approved):

PR #528, opened 2022-01-26
QPPA-6005: update measure strata for 366 - 2022

ckawell:
Changes the name for the second strata for measure 366. Also updates a couple other strata name to be less susceptible to auto Excel date formatting. 

* https://jira.cms.gov/browse/QPPA-6005


michelingle (changes requested):
Please reload the updated file in QPPA-6005 as 2 of the strata were auto-converted to dates by Excel...

john-manack (approved):

PR #538, opened 2022-02-10
QPPA-6037, QPPA-6038 Suppress 111 eCQM and remove the benchmark 2021

ckawell:
This ticket changes Y2021 111 eCQM truncation to a suppression in the measures repo and removes the benchmark.

* https://jira.cms.gov/browse/QPPA-6037
* https://jira.cms.gov/browse/QPPA-6038


michelingle (changes requested):
Hey, this may not need to be fixed but why are we adding in a submission method for CAHPSSurveyVendor?  The only submissionMethod we have for 321 (CAHPS) is certifiedSurveyVendor...

michelingle (approved):
Following up with RAND as to whether the change was intentional and will be the new term for future years.

PR #556, opened 2022-04-14
QPPA-6097: Update schemas

ckawell:
During the Data Dictionary Audit, some measure and benchmark fields were found to be lacking descriptions in their respective schemas. This PR adds descriptions for those fields.

* https://jira.cms.gov/browse/QPPA-6097

### DIFF
--- a/benchmarks/2022/benchmarks-schema.yaml
+++ b/benchmarks/2022/benchmarks-schema.yaml
@@ -33,5 +33,6 @@ definitions:
         minItems: 9
         maxItems: 10
       submissionMethod:
+        description: The method for submitting the measure performance data to which this benchmark applies.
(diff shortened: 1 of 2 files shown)


michelingle (changes requested):
line 217: Change "Used to truncate submissions data" to "Used to indicate that submissions data should be truncated" ...
line 221: There should be some part of benchmarks that calls out "removed".  Note for next year, we will add back "flat".  So there will be two options: flat, removed.

michelingle (changes requested):
Line 155: change to similar description as line 213 but change the end to "... listed in clinicalGuidelinesChanged"

michelingle (approved):

nicholas-gates (approved):

PR #560, opened 2022-05-12
QPPSF-10173 - PY21 Benchmarks Two Decimal Places

KyleApfel:
#### Motivation for change

Reducing the precision of performance benchmarks from 4 decimal places to 2. This is surrounding policy decisions to determine if this is the path we want to go forward. 

Because of the nature of AC measures being decimal based scores, these must stay at 4 decimal places to be statistically relevant. It was also discovered these AC measures were wrongly trimmed to 9 decimal rather than 10 which is required. 

#### Release checklist:

* [ ] Documentation updated
* [ ] Unit tests added/passing

* https://jira.cms.gov/browse/QPPSF-10173


Jpec07 (approved):
:taco:

shadford (approved):

saquino0827 (approved):

elanperry (approved):

michelingle (approved):

PR #614, opened 2022-10-24
QPPA-6963 Update packages that use minimatch

ljackson-sb:
#### Motivation for change

SFUI team reported a Snyk error for minimatch v3.0.4 in their PR checker. You can view more about the vulnerability [here](https://security.snyk.io/vuln/SNYK-JS-MINIMATCH-3050818).

#### What is being changed

We are not using the minimatch package directly, but several of the packages we use depend on minimatch. The packages that have updated minimatch beyond 3.0.4 have been updated (ESLint and Mocha). Unfortunately many of the nested dependency packages are still using v3.0.4:

- [`config-array`](https://github.com/humanwhocodes/config-array/blob/main/package.json#L47) (used by ESLint)
- [`eslint-config-nava`](https://github.com/navapbc/eslint-config-nava/blob/main/package.json#L24) (using an older version of [`eslint-plugin-import`](https://github.com/import-js/eslint-plugin-import/blob/v2.25.4/package.json#L111))
- [`eslint-plugin-node`](https://github.com/mysticatea/eslint-plugin-node/blob/master/package.json#L20)
- [`test-exclude`](https://github.com/istanbuljs/test-exclude/blob/master/package.json#L35)

So I'm not sure if this will still cause a Snyk error in their build. We can monitor these packages and/or submit issues requesting they update this package.

#### Release checklist:

* [ ] Documentation updated
* [ ] Unit tests added/passing

* https://jira.cms.gov/browse/QPPA-6963


package-lock.json
```diff
@@ -1,8 +1,10586 @@
 {
   "name": "qpp-measures-data",
   "version": "5.5.7",
-  "lockfileVersion": 1,
+  "lockfileVersion": 2,
```
ckawell-sb, line 4: I don't think we're ready to upgrade to version 2. Should be able to fix this with running `npm i` with the current npm version for Sub API (`npm use v14.19` I think)

ckawell-sb (changes requested):

package-lock.json
ljackson-sb: Good catch! Fixed

nicholas-gates (approved):

PR #637, opened 2022-11-30
QPPA-7169: misc script updates

ckawell-sb:
Misc. Measures Scripts upgrades

Features and bug fixes:

- Added the ability to test the ingestion process without persisting via the -t flag.
- Removed warning message when running the export script with an existing `/misc` directory and sub directories.
- Added check to see if new Exclusions and Substitutes exist. 
    - Throws error and fails when **updating** a measure with a bad Exc. or Sub.
    - Returns warning and passes when **adding** new measure with a bad Exc. or Sub.
- Fixed bug where a row with missing trailing commas would silently not be processed.
- Ingestion script now auto removes trailing empty rows from CSV. This step is no longer needed to be performed manually.
- Renamed the export script to be consistent with naming convention.
- Updated README

* https://jira.cms.gov/browse/QPPA-7169


scripts/measures/2023/update-measures-util.ts
```diff
@@ -132,8 +128,18 @@ export function updateMeasuresWithChangeFile(
             exports.deleteMeasure(acceptedDeleteStack[i].measureId, acceptedDeleteStack[i].category, measuresJson);
         }
 
-        exports.updateChangeLog(fileName, changesPath);
-        info(`File '${fileName}' successfully ingested into measures-data ${performanceYear}`);
+        //ingest changed measures
```
ckawell-sb, line 131: Moved changes to after additions and deletions so we can have the very latest version of the measures data before validating changes.

scripts/measures/export-measures
```diff
@@ -14,8 +14,8 @@ if (( $currentPerformanceYear >= 2018 )) && (( $currentPerformanceYear <= $maxPe
   tsc -p .
 
   echo -e "Converting to CSV..."
-  mkdir tmp/
-  mkdir tmp/$currentPerformanceYear
+  mkdir -p tmp/
+  mkdir -p tmp/$currentPerformanceYear
```
ckawell-sb, line 18: Now won't throw message if directories already exist.

scripts/measures/lib/csv-json-converter.ts
```diff
@@ -56,6 +56,11 @@ export function convertCsvToJson(csv: any) {
 
         return measure;
     });
+
+    //remove any empty rows
+    _.remove(mappedCsv, _.isEmpty);
+    
+    return mappedCsv;
```
ckawell-sb, line 63: Now no need to manually delete ',,,,,,,,,,...' trailing csv rows.

john-manack (approved):

ljackson-sb (approved):

PR #657, opened 2023-01-09
QPPSF-11471: Create 2023 benchmarks file

ckawell-sb:
- New PY2023 Benchmarks File added

- Scripts updated to handle new CSV format


benchmarks/2023/benchmarks-schema.yaml
```diff
@@ -0,0 +1,84 @@
+$id: https://github.com/CMSgov/qpp-measures-data/versions/0.0.1/benchmarks-schema.yaml
```
ckawell-sb, line 1: This Schema file is a copy-paste from PY22. Is this what it should be?

ljackson-sb (approved):

scripts/benchmarks/parse-benchmarks-data.js
```diff
@@ -86,7 +99,7 @@ if (benchmarkYear && performanceYear) {
   process.stdin.on('end', function() {
     // Quote option update to handle 2020 data
     const columns = benchmarkType === 'MCC' ? MCC_BENCHMARK_CSV_COLUMNS : BENCHMARK_CSV_COLUMNS;
-    parse(benchmarksData, {columns, from: 3, quote: '"'}, function(err, records) {
+    parse(benchmarksData, {columns, from: 2, quote: '"'}, function(err, records) {
```
ljackson-sb, line 102: what is the reason for this change?

ckawell-sb: I'm guessing in earlier years there were two lines of headers. But with this CSV, the data starts on row 2. I was seeing the first row being ignored until I changed that.

ljackson-sb (dismissed):

benchmarks/2023/benchmarks-schema.yaml
pkasarski: Yes, updates had been added there.

pkasarski (approved):

ljackson-sb (approved):

PR #663, opened 2023-01-26
QPPA-7260: Clean up benchmarks scripts to support 2023 and 2022 data

john-manack:
#### Motivation for change

- failing tests needed to pass by updating the scripts to process with both 2022 and 2023 csv files

#### Release checklist:

* [ ] Documentation updated

* https://jira.cms.gov/browse/QPPA-7260

### DIFF
--- a/scripts/benchmarks/parse-benchmarks-data.js
+++ b/scripts/benchmarks/parse-benchmarks-data.js
@@ -14,8 +14,6 @@ const BENCHMARK_CSV_COLUMNS = [
   'benchmark',
   'standardDeviation',
   'average',
-  'decile1',
-  'decile2',
   'decile3',
   'decile4',
   'decile5',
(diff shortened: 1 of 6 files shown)


scripts/benchmarks/parse-benchmarks-data.js
```diff
@@ -78,12 +65,29 @@ if (performanceYear >= 2020) {
   MCC_BENCHMARK_CSV_COLUMNS.push('isHighPriority');
 }
 
-// New 2022 data update
-if (performanceYear >= 2022) {
+// New 2023 data update
+if (performanceYear >= 2023) {
   BENCHMARK_CSV_COLUMNS.push('isInverse');
```
john-manack, line 70: @peg-uehling Chris and I made the decision to remove `isInverse` and `metricType` from the expected csv columns for 2022 data only. When the script expects those columns for 2022, it breaks when running the current csv containing 2022 data. If we end up loading new 2022 data with these fields, we will get a helpful developer error, and we can simply add-back those columns here.

john-manack: It these fields are supposed to be option for 2022 according to this ticket:
https://jira.cms.gov/browse/QPPA-6971

scripts/benchmarks/parse-benchmarks-data.js
```diff
@@ -14,8 +14,6 @@ const BENCHMARK_CSV_COLUMNS = [
   'benchmark',
   'standardDeviation',
   'average',
-  'decile1',
```
ckawell-sb, line 17: Why are we removing these new columns here?

peg-uehling (approved):

ljackson-sb (approved):

